PR #10, opened 2015-12-16
Added gitVersionDescriber() command to allow settings

jggc:
The idea is to allow the script to use the equivalent of "git describe --long" by getting the version describer and allowing the script to do whatever it wants with it.

It was necessary to go through some refactoring of GitVersionPlugin.groovy to allow this to happen.
I added a test case and refactored repetitive git commands to private methods.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
@@ -43,9 +43,11 @@ class GitVersionPlugin implements Plugin<Project> {
     }
 
     private String doDescribe(Project project) {
+
+        Git git = ensureGit(project);
         try {
             String version = ensureDescribeCommand(project).call() ?: UNSPECIFIED_VERSION
-            boolean isClean = ensureGit(project).status().call().isClean()
+            boolean isClean = git.status().call().isClean()
             return version + (isClean ? '' : '.dirty')
 
         } catch (Throwable t) {
@@ -55,7 +57,7 @@ class GitVersionPlugin implements Plugin<Project> {
 
     private DescribeCommand ensureDescribeCommand(Project project) {
         if (describeCommand == null) {
-            Git git = ensureGit()
+            Git git = ensureGit(project)
             describeCommand = git.describe()
         }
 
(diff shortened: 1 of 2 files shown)


src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -151,6 +141,45 @@ class GitVersionPluginTests extends Specification {
         buildResult.standardOutput.contains(':printVersion\n1.0.0.dirty\n')
     }
 
+    def 'git describe long when annotated tag is present and long is true' () {
+        given:
+        buildFile << '''
+            plugins {
+                id 'com.palantir.git-version'
+            }
+            gitVersionDescriber().setLong(true)
```
markelliot, line 150: I think the mutability here is pretty undesirable, as some invocations may make it much harder (especially in multi project builds) to keep calls to `gitVersion()` idempotent.

markelliot:
I'm not sure the additional complexity here is a net improvement. I suspect it would be better to add a config block that allows tuning of the describe command, and can imagine going as far as including a lambda that takes a `Git` instance and returns a version, but in current form doesn't quite work.

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -24,27 +25,57 @@ class GitVersionPlugin implements Plugin<Project> {
...
     private static final String UNSPECIFIED_VERSION = 'unspecified'
+    private DescribeCommand describeCommand = null
+    private Git git;
 
     void apply(Project project) {
         project.ext.gitVersion = {
+            doDescribe(project)
+        }
+
+        project.ext.gitVersionDescriber = {
```
markelliot, line 36: This should really be part of a configuration block, and the lambda should follow a prototype of `(Git) -> String`.

jggc:
@markelliot I get your points and agree. When you say configuration block are you thinking of something at the top of the project's build file that configures the instance of gitVersion or just adding an optional closure parameter to the gitVersion function call?

I think the latter is preferable, I don't know how to do that just yet but I'll check and update that p-r once I'm done. Thanks for reviewing my work.

markelliot:
I was tempted to suggest just taking a closure as an argument to `gitVersion`, but it might be more error prone than taking a configuration.

In the end, it might make sense, at least for now, to just make a `gitLongVersion()` -- one of the reasons this plugin exists is to reduce the variation in version styles.

PR #15, opened 2015-12-29
Describe options - Another try

jggc:
This time I added two optional parameters to the gitVersion closure. I think it's much better than the other version but I don't really like that we need to specify the first parameter when we want to set the target.

I thought it was possible to give the parameter name in the function call like in python but it seems there are some limitations to that.

I guess there is some groovy trick that I don't know to do that.


markelliot:
Thanks for this, but I think we'll try to take another pass at it in the near future but expect we'll add another alternative function call (rather than modifying `gitVersion`) to get the long version (maybe `gitLongVersion()`) to keep things simple. Will comment on your open issue when we get to it.

PR #18, opened 2016-02-02
Print "unspecified" instead of throwing error when .git directory is missing

tomshen:


markelliot:
Oh I see, this did have a test. Why is it valid to ask for a version when there's no git repo? (the idea was that this was good signal you should have a git repo, I think)

tomshen:
We develop on template projects that are inside a larger git repo. So, each template project lacks a .git directory. We could generate projects from these templates, which would have .git directories, but this makes the development process a lot more annoying.

markelliot:
I would expect that you only apply this plugin to your rootProject and not to individual sub-projects. You can use the `allprojects` block in Gradle to apply the version to everything. e.g. most of my multi-project setups are:

```
allproject {
    version gitVersion()
    group 'com.palantir.<group>'
}
```

Does that not work for you here?

tomshen:
We're developing on templates for a Yeoman generator. Each template is meant to be a distinct root project. So this wouldn't work, since the templates don't share a root project.

markelliot:
For tests, you could initialize a git repo; I guess I'm unsure what other scenario you'd be running those projects (for non-tests?)

jmcampanini:
i believe the PR linked above should address our concerns

PR #20, opened 2016-02-23
Add test coverage for symlink behavior

markelliot:


punya:
:+1: seems worth merging as a guard against future regressions.

PR #24, opened 2016-03-02
add functionality to search parent dirs for git repo

### LINKED ISSUE #23: Search parent directories for .git
The plugin currently expects the .git directory to be in the project root, but git itself will search up through parent directories.

jmcampanini:
fixes #23 
closes #18

### DIFF
--- a/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
@@ -27,7 +27,7 @@ class GitVersionPlugin implements Plugin<Project> {
 
     void apply(Project project) {
         project.ext.gitVersion = {
-            File gitDir = new File(project.rootDir, '.git')
+            File gitDir = findRootGitDir(project.rootDir)
             if (!gitDir.exists()) {
                 throw new IllegalArgumentException('Cannot find \'.git\' directory')
             }
@@ -46,5 +46,20 @@ class GitVersionPlugin implements Plugin<Project> {
             println project.version
         }
     }
-}
 
+    private File findRootGitDir(File currentRoot) {
+        File gitDir = new File(currentRoot, '.git')
+
+        if (gitDir.exists()) {
+            return gitDir
+        }
+
+        // stop at the root directory, return non-existing File object
+        if (currentRoot.parentFile == null) {
+            return gitDir
+        }
+
+        // look in parent directory
+        return findRootGitDir(currentRoot.parentFile)
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -46,5 +46,20 @@ class GitVersionPlugin implements Plugin<Project> {
...
 
+    private File findRootGitDir(File currentRoot) {
+        File gitDir = new File(currentRoot, '.git')
+
+        if (gitDir.exists()) {
+            return gitDir
+        }
+
+        // do not follow to the root directories, return non-existing File object
+        if (currentRoot.parent == '/' || currentRoot.parent == '~') {
```
markelliot: does this work on windows?

sfackler: File::getParentFile will return null if there is no parent. That might be a more robust check.

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -46,5 +46,20 @@ class GitVersionPlugin implements Plugin<Project> {
...
 
+    private File findRootGitDir(File currentRoot) {
+        File gitDir = new File(currentRoot, '.git')
+
+        if (gitDir.exists()) {
+            return gitDir
+        }
+
+        // stop at the root directory, return non-existing File object
+        if (currentRoot.parentFile == null) {
```
jmcampanini, line 58: 1. switched to using `File` methods
2. detecting root by using `File::parentFile == null`

this should be safe on windows. don't have a win machine to test on though :-\

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -52,6 +52,27 @@ class GitVersionPluginTests extends Specification {
...
+            version gitVersion()
+        '''.stripIndent()
+        Git git = Git.init().setDirectory(rootFolder).call();
+
+        when:
+        // will build the project at projectDir
+        BuildResult buildResult = with('printVersion').build()
+
+        then:
+        buildResult.output.contains(':printVersion\nunspecified\n')
```
markelliot: sorry I didn't catch this earlier, would like to actually have a commit to test we get a real version (as that guarantees there's actually a git folder there)

jmcampanini:
let me know if this needs to be squashed before merging

markelliot:
yep, please squash and then I'll merge + release.

jmcampanini:
squashed :-D

PR #35, opened 2016-06-11
Update plugin

nmiyake:


nmiyake:
@markelliot (low-pri, but for SA)

PR #38, opened 2016-09-26
return dummy VersionDetails object when version is unspecified

qinfchen:
When version is unspecified, user will get an error when trying to obtain properties from VersionDetails

```
details.lastTag
details.commitDistance
details.gitHash
```

instead of checking for null before accessing these properties, can we just return a dummy object?

### DIFF
--- a/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
@@ -71,11 +71,7 @@ class GitVersionPlugin implements Plugin<Project> {
             String description = gitDesc(project)
             String hash = gitHash(project)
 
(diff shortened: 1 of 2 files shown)


qinfchen:
@zoubeiri @tomkozlowski

markelliot:
This is a decent API break. Could you elaborate on the use case where this is necessary?

qinfchen:
This is when the project is newly created and without a tag, you would get a NPE when trying to access any properties from VersionDetails. I am not sure if there are other cases, but you can't directly access `lastTag`, `commitDistance` and `gitHash` without checking for null.

tomkozlowski:
This is in reference to a regression (commit # dc8fb8f289b62714cc6b2825ac97dc59dab57041) we had on template-sls when switching to `gradle-git-version` from `flex-version`

delta003:
Closing this since no activity for a while

PR #43, opened 2016-12-30
Add VersionDetails#branchName for current branch of HEAD

### LINKED ISSUE #42: Add branchName to VersionDetails
Which is an `optional<string>`

jared2501:
Fixes #42.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -10,26 +10,33 @@ It behaves exactly as the JGit `git describe` method behaves, except that when t
 state, appends `.dirty` to the version string.
 
 The second, called `versionDetails()`, returns an object containing the specific details of the version string:
-the tag name, and the commit count since the tag.
+the tag name, the commit count since the tag, the current commit hash of HEAD, and an optional branch name of HEAD.
 
 Usage
 -----
 Apply the plugin using standard Gradle convention:
 
-    plugins {
-        id 'com.palantir.git-version' version '<current version>'
-    }
+```groovy
+plugins {
+    id 'com.palantir.git-version' version '<current version>'
+}
+```
 
 Set the version of a project by calling:
 
-    version gitVersion()
+```groovy
+version gitVersion()
+```
 
 You can get an object containing more detailed information by calling:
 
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -10,26 +10,33 @@ It behaves exactly as the JGit `git describe` method behaves, except that when t
...
-    def details = versionDetails()
-    details.lastTag
-    details.commitDistance
-    details.gitHash
+```groovy
+def details = versionDetails()
+details.lastTag
+details.commitDistance
+details.gitHash
+details.branchName // is null if the repository in detached HEAD mode
```
jared2501, line 38: made the branchName null instead of Optional since it seemed more groovy to do so...

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -62,6 +63,21 @@ class GitVersionPlugin implements Plugin<Project> {
...
+        Git git = gitRepo(project)
+        try {
+            def ref = git.repository.getRef(git.repository.branch)
+            if (ref == null) {
+                return null
+            } else {
+                return ref.getName().substring(Constants.R_HEADS.length())
+            }
+        } catch (Throwable t) {
+            return UNSPECIFIED_VERSION
```
markelliot: Do you have a test that covers hitting this case?

jared2501: hmm, no I don't (just copied + pasted it from other methods). Looking a bit closer at it, it will only be hit in weird exception cases (like a malformed git repo).. Removed this try/catch.

jared2501: If https://github.com/palantir/gradle-git-version/pull/38 merges then this behaviour will need to change obviously.

jared2501:
@markelliot - addressed comments

markelliot (approved):

PR #44, opened 2017-02-07
Log warning when no tags found in git repo

### LINKED ISSUE #40: gitHash returns 'unspecified'
I have a git repository. At the moment it has no tags. I would still expect git hash to return the HEAD object hash, but instead it returns 'unspecified'.

hxu:
Fixes #40


delta003:
Fixed in new implementation

PR #48, opened 2017-03-08
Match output of 'git describe --first-parent'

### LINKED ISSUE #46: Change plugin to match behavior of "git describe --first-parent"
Currently, the output of `gradle-git-version` matches the output of `git describe`. Although this is fine in the common case, it's often not the desired behavior if merge commits exist.

For example, consider the following case:

* Mainline branch has some tag (`1.0.0-dev`) and development is proceeding normally
* Some other branch (for example, `1.0.0-hotfix`) is merged into the mainline branch
* `gradle-git-version` is run on the mainline branch

In this scenario, we typically want the output to still be `1.0.0-dev-*`. However, because `git describe` returns the nearest ancestor tag, the merge causes `1.0.0-hotfix` to be found as a parent tag that is "closer" to the commit than `1.0.0-dev`, so you get `1.0.0-hotfix-*` instead.

The fix would be to match the behavior of `git describe --first-parent` instead -- this command only considers the first parent commit for merge commits, and would cause this to work correctly. I believe that this is almost always the desired behavior.

nmiyake:
Makes it so that, if a merge causes a tag to be added to the
history, the tags from the merge will not be used as part of
the output for the description.

Fixes #46

### DIFF
--- a/src/main/groovy/com/palantir/gradle/gitversion/DescribeFirstParentCommand.java
+++ b/src/main/groovy/com/palantir/gradle/gitversion/DescribeFirstParentCommand.java
@@ -1,0 +1,364 @@
+/*
+ * Copyright 2017 Palantir Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * <http://www.apache.org/licenses/LICENSE-2.0>
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Derived from: https://github.com/eclipse/jgit/blob/3b4448637fbb9d74e0c9d44048ba76bb7c1214ce/org.eclipse.jgit/src/org/eclipse/jgit/api/DescribeCommand.java
+ *
+ * Copyright (C) 2013, CloudBees, Inc.
+ * and other copyright owners as documented in the project's IP log.
+ *
+ * This program and the accompanying materials are made available
+ * under the terms of the Eclipse Distribution License v1.0 which
+ * accompanies this distribution, is reproduced below, and is
+ * available at http://www.eclipse.org/org/documents/edl-v10.php
+ *
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or
+ * without modification, are permitted provided that the following
+ * conditions are met:
+ *
+ * - Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * - Redistributions in binary form must reproduce the above
+ *   copyright notice, this list of conditions and the following
+ *   disclaimer in the documentation and/or other materials provided
+ *   with the distribution.
+ *
+ * - Neither the name of the Eclipse Foundation, Inc. nor the
+ *   names of its contributors may be used to endorse or promote
+ *   products derived from this software without specific prior
+ *   written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND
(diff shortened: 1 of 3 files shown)


nmiyake:
@uschi2000 @markelliot

nmiyake:
Ping for feedback

src/main/groovy/com/palantir/gradle/gitversion/DescribeFirstParentCommand.java
```diff
@@ -0,0 +1,304 @@
+package com.palantir.gradle.gitversion;
```
uschi2000, line 61: needs copyright and link to origin

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -44,7 +45,10 @@ class GitVersionPlugin implements Plugin<Project> {
     private String gitDesc(Project project) {
         Git git = gitRepo(project)
         try {
-            String version = git.describe().call() ?: UNSPECIFIED_VERSION
+            DescribeFirstParentCommand describe = new DescribeFirstParentCommand(git.getRepository())
+//            DescribeCommand describe = new DescribeCommand(git.getRepository())
```
uschi2000: huh?

nmiyake: Oops sorry poor form on my part -- debugging code that should be removed.

src/main/groovy/com/palantir/gradle/gitversion/DescribeFirstParentCommand.java
nmiyake: Yup had this on my todo list but forgot to do before push -- will do

uschi2000 (approved):
i'm fine with this, but sort of think that relying on OS `git` is simpler?

nmiyake:
Yup, just a matter of trade-offs. If we depend on OS git, we'll need to assume that the host system has `git` installed, that Gradle can call it using `git` and that it will resolve to the correct command, and that all of the versions of `git` we expect to use are of a version that has the commands/flags that we expect. It will also be a bigger code change for this Gradle plugin.

I don't have a strong preference either way. Even though this change does add a bit of code, I think that the surface area is small (and targeted) enough that I think it makes sense. We can also still migrate to shelling out later if we decide that we want to do that (or find more Git features that we want to use).

If you would prefer that we use shell out to Git, happy to open a PR that converts this plugin to do that. However, for now I would lean towards adding this functionality in this manner in JGit first.

nmiyake:
Addressed PR feedback

uschi2000:
+1

nmiyake:
Thanks! @markelliot are you good with this change?

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -155,6 +156,40 @@ class GitVersionPluginTests extends Specification {
...
+        // switch back to main branch and merge hotfix branch into main branch
+        git.checkout().setName(master).call()
+        git.merge().include( git.getRepository().getRef("hotfix")).setFastForward(MergeCommand.FastForwardMode.NO_FF).setMessage("merge commit").call()
+
+        when:
+        BuildResult buildResult = with('printVersion').build()
+
+        then:
+        buildResult.output =~ ":printVersion\n1.0.0-1-g[a-z0-9]{7}\n"
+    }
```
markelliot, line 193: I'd like to see the same test but also tagging a 2.0.0 after the hotfix merges to ensure we see that come up.

nmiyake: Done

nmiyake:
Addressed PR feedback

markelliot (approved):

nmiyake:
@markelliot @uschi2000 can we tag a 0.5.4 or 0.6.0 to publish a release that includes this change?

uschi2000:
I don't have permissions, sorry.

markelliot:
@uschi2000 you should now

nmiyake:
Ping

uschi2000:
released as 0.6.0

nmiyake:
Thanks!

PR #50, opened 2017-03-13
change .dirty to -dirty

derenrich:
This makes versions created by this plugin fully compatible with the SLS spec (I'd link but that repo is no longer public). This commit is a result of a discussion on the SLS spec.

Probably should bump the major version of this plugin after this commit since it could break people's workflows.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
@@ -47,7 +47,7 @@ class GitVersionPlugin implements Plugin<Project> {
             DescribeFirstParentCommand describe = new DescribeFirstParentCommand(git.getRepository())
             String version = describe.call() ?: UNSPECIFIED_VERSION
             boolean isClean = git.status().call().isClean()
-            return version + (isClean ? '' : '.dirty')
+            return version + (isClean ? '' : '-dirty')
         } catch (Throwable t) {
             return UNSPECIFIED_VERSION
         }
--- a/src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
@@ -134,7 +134,7 @@ class GitVersionPluginTests extends Specification {
         BuildResult buildResult = with('printVersion').build()
 
         then:
-        buildResult.output.contains(':printVersion\nunspecified.dirty\n')
+        buildResult.output.contains(':printVersion\nunspecified-dirty\n')
     }
 
     def 'git describe when annotated tag is present' () {
@@ -248,7 +248,7 @@ class GitVersionPluginTests extends Specification {
         BuildResult buildResult = with('printVersion').build()
 
         then:
-        buildResult.output.contains(':printVersion\n1.0.0.dirty\n')
+        buildResult.output.contains(':printVersion\n1.0.0-dirty\n')
     }
 
     def 'git describe and clean when symlink is present' () {


ilyanep:
+2 lgtm

uschi2000:
sgtm

derenrich:
who presses the merge button and cuts the release if not you?

markelliot:
Sorry, I'm confused, why do we want dirty versions to be valid versions in our deployability ecosystem?

derenrich:
This ticket is not asking for support for dirty versions. We already have support explicitly called out. See for example [test cases in sls-packaging](https://github.com/palantir/sls-packaging/blob/338968099c704088ef8d8686fabaf128e0f9839d/sls-versions/src/test/java/com/palantir/slspackaging/versions/SlsProductVersionsTest.java#L32). The spec for the manifest specifically calls out dirty version strings as valid. 

I have reasons for wanting this. They don't involve publishing. They involve local testing. 

Currently this tool is incompatible with our other tools since it produces versions that are not valid SLS manifest specifications when it could otherwise easily avoid that.

markelliot:
sls-packaging does, in fact, allow versions that end in ".dirty", allowing you to build local artifacts for local testing. The manifest spec does not allow these versions and shouldn't allow these versions.

If you only plan to do local testing, why is this breaking change required?

On Mar 13, 2017, at 8:24 PM, Daniel Erenrich <notifications@github.com<mailto:notifications@github.com>> wrote:


This ticket is not asking for support. We already have support explicitly called out. See for example test cases in sls-packaging[github.com]<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_palantir_sls-2Dpackaging_blob_338968099c704088ef8d8686fabaf128e0f9839d_sls-2Dversions_src_test_java_com_palantir_slspackaging_versions_SlsProductVersionsTest.java-23L32&d=DwMFaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=Vp81aOxWZvuVgVK_wp-VF3pIYG92B19LcCw6XKeYC0U&m=3KihTmQHOgYtaxI840tvZKUYfdJNzkCULLU1t28oZSM&s=TzfMoOT_TINkPKklrMCLsf7VfiDepX4qhepPMM3WS0s&e=>. The spec for the manifest specifically calls out dirty version strings as valid.

I have reasons for wanting this. They don't involve publishing. They involve local testing.

Currently this tool is incompatible with our other tools since it produces versions that are not valid SLS manifest specifications when it could otherwise easily avoid that.

derenrich:
I really didn't want to go into this level of detail but ok here's the full story.

You are actually mistaken. sls-packaging allows versions that end in `.dirty` but only if that is proceeded by a dash. That is per the SLS spec (again there's a ticket where discussion occurred but I can't link to it). If you use this gradle plugin and are exactly on a tag then the version string will be `1.2.3.dirty` and so there won't be a dash before the `.dirty` and so the version is invalid. That is why I am making this change.

Also the manifest spec does allow these version? They are unordered versions which I understand are allowed. They are explicitly mentioned in the spec as valid.

uschi2000:
Dan, we can change the non-orderable format to something like `\d+\.\d+\.\d+(-[a-z0-9-.]*)?(\.dirty)?`

uschi2000:
fyi @derenrich  https://github.com/palantir/sls-packaging/pull/202

derenrich:
Ok but then the sls-packaging gradle plugin will no longer actually be compatible with the actual spec?  Or are you proposing changing the SLS spec also?

uschi2000:
yep, changing the spec too

derenrich:
Just FYI. I started this journey at the spec asking it to be changed. It was suggested this was the easier change.

markelliot:
It's unreasonable that `1.0.0-dirty` is allowable but `1.0.0.dirty` is not. I'd prefer we make the spec more pedantic w/r/t dirty markers instead of break this plugin.

derenrich:
Ok. Maybe you can convince them better than I.

PR #52, opened 2017-03-28
Use 'git' command to generate description string

nmiyake:


nmiyake:
markelliot @uschi2000 for consideration...

Per Rob's comment preceding https://github.com/palantir/gradle-git-version/pull/48#issuecomment-285535448, this shells out to the local `git` command to generate the "describe" string.

Change motivated by the fact that the native JGit `--first-parent` implementation I submitted did not seem to work in a more complicated repository (specifically, version string produced by `gradle-git-version` differs from that produced by `git describe --tags  --first-parent` in one of our internal repositories). Unclear as to if this is due to implementation detail of JGit, bug in my implementation, or something else. This change makes the operation that is being performed explicit and consistent across our build tools.

If you would prefer not to accept this PR, then I will likely request to revert my previous `--first-parent` change since I'm unable to fully explain the behavior, which is unsettling to me. In that case, I'll likely just live with the current constraints of not using `--first-parent` (undesired output in certain gitflow scenarios) or submit a PR where this `--first-parent` behavior can be turned on using a flag or argument to the `version()` function when called from a Gradle script

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -44,8 +45,11 @@ class GitVersionPlugin implements Plugin<Project> {
     private String gitDesc(Project project) {
         Git git = gitRepo(project)
         try {
-            DescribeFirstParentCommand describe = new DescribeFirstParentCommand(git.getRepository())
-            String version = describe.call() ?: UNSPECIFIED_VERSION
+            // back-compat: the JGit "describe" command throws an exception in repositories with no commits, so call it
+            // first to preserve this behavior in cases where this call would fail but native "git" call does not.
+            new DescribeCommand(git.getRepository()).call()
+
+            String version = runGitCommand(project.rootDir, "describe", "--tags", "--first-parent") ?: UNSPECIFIED_VERSION
```
markelliot, line 55: pretty sure we didn't have to strip a line separator from jgit's output, but likely we need to here?

nmiyake: Already done as part of the `runGitCommand` method at line 59:

```java
return builder.toString().trim()
```

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -128,4 +132,31 @@ class GitVersionPlugin implements Plugin<Project> {
...
+        StringBuilder builder = new StringBuilder()
+        String line = null
+        while ((line = reader.readLine()) != null) {
+            builder.append(line)
+            builder.append(System.getProperty("line.separator"))
+        }
+
+        int exitCode = process.waitFor()
+        if (exitCode != 0) {
+            return ""
```
uschi2000, line 166: we need to alert users when the `git` command isn't available.

uschi2000 (changes requested):
need to alert when git command is not available

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
nmiyake: Makes sense.

In that case, would you prefer that we simply throw an exception, or log a warning and fall back to JGit `git describe` behavior? Presume the former, but wanted to check.

uschi2000: The main motivation for this PR is to cut code, so yes, the former.

uschi2000: Pondering ... how stable is the `git tag --describe` behavior across different `git` versions that we may expect to be installed in places?

uschi2000: (we control with jgit behavior via the dependency version, but we cannot control the installed git behavior ... unless we gate on a particular version (range) being available.)

nmiyake: Yup, that is a totally valid concern.

Although I don't have concrete data, I'm fairly confident that the `git --describe --tag` behavior should be stable -- I think it would be a major back-compat break on their end if it was not.

The `describe` command and `--tags` flag have been around for as long as I can find online. The `--first-parent` flag was introduced in Git 1.8.4, which was released 08/23/2013.

The CircleCI nodes have version 2.7.4 of `git`, while locally I have 2.6.3. I think it's a fairly safe assumption that people have a version of Git that supports this.

The other option here is to make `--first-parent` an opt-in behavior, and to only use native `git` if we want that behavior. However, my other contention is that other tooling that wants to produce similar version strings will most likely shell out to `git` to do so, so in a way I would rather tie the behavior of this output to the output of `git` rather than to the particulars of the implementation of JGit.

nmiyake: We can still discuss the broader issue, but added check for `git` executable. If it does not exist, throws an exception of the following form (this is from unit test output when I changed the check to check for an executable named `gitz`, which obviously does not exist):

```
|           * What went wrong:
|           A problem occurred evaluating root project 'junit2262689230640608264'.
|           > Cannot run program "gitz": error=2, No such file or directory
```

uschi2000 (approved):
lgtm

nmiyake:
Thanks! If I could get another release so I can apply this change that would be awesome :)

PR #58, opened 2017-05-10
Add VersionDetails#isCleanTag

### LINKED ISSUE #57: FR: isRelease boolean on VersionDetails
I frequently see logic in other gradle files that is conditional on whether the project is 'released' or still 'under development':

```groovy
def isRelease = (project.version.toString() ==~ /[0-9]+(\.[0-9]+)+((-(beta|rc)[0-9]{1,2})(\.[0-9])?)?/)
dockerPush.onlyIf { isRelease }
```

To avoid duplicating this regex, let's add a convenience boolean to the VersionDetails object, allowing us to write: 

```groovy
dockerPush.onlyIf { gitVersion().isRelease }
```

iamdanfox:
fixes #57 

Instead of the common boilerplate:

```groovy
def isRelease = (project.version.toString() ==~ /[0-9]+(\.[0-9]+)+((-(beta|rc)[0-9]{1,2})(\.[0-9])?)?/)
dockerPush.onlyIf { isRelease }
```

You can now write:

```groovy
dockerPush.onlyIf { gitVersion().isCleanTag }
```

A tiny refactor was necessary to do this in a non-nasty way... the `VersionDetails` class now has some getters that derive information from the raw info we get from git.

Wasn't 100% sure what to call the field... `isRelease` seemed a bit presumptuous (people might have funny tags like 1.2.3-dev), so I figured `isCleanTag` was the minimal description...

What do you think @uschi2000 / @markelliot?

### DIFF
--- a/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
@@ -20,6 +20,7 @@ import org.eclipse.jgit.api.DescribeCommand
 import org.eclipse.jgit.api.Git
 import org.eclipse.jgit.internal.storage.file.FileRepository
 import org.eclipse.jgit.lib.Constants
+import org.eclipse.jgit.lib.ObjectId
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 
@@ -61,7 +62,11 @@ class GitVersionPlugin implements Plugin<Project> {
     @Memoized
     private String gitHash(Project project) {
         Git git = gitRepo(project)
-        return git.getRepository().getRef("HEAD").getObjectId().abbreviate(VERSION_ABBR_LENGTH).name()
+        ObjectId objectId = git.getRepository().getRef("HEAD").getObjectId();
+        if (objectId == null) {
+            return null
+        }
+        return objectId.abbreviate(VERSION_ABBR_LENGTH).name()
     }
 
     @Memoized
@@ -78,10 +83,6 @@ class GitVersionPlugin implements Plugin<Project> {
     @Memoized
     private VersionDetails versionDetails(Project project) {
         String description = gitDescribe(project)
-        if (description.equals(UNSPECIFIED_VERSION)) {
-            return null
-        }
-
         String hash = gitHash(project)
         String branchName = gitBranchName(project)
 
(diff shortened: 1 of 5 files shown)


src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -283,7 +283,7 @@ class GitVersionPluginTests extends Specification {
         buildResult.output.contains(':printVersion\n1.0.0\n')
     }
 
-    def 'version details null when no tags are present' () {
+    def 'version details not null when no tags are present' () {
```
iamdanfox, line 286: Note: this is a behaviour change

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -20,153 +20,89 @@ import org.eclipse.jgit.api.DescribeCommand
...
             // first to preserve this behavior in cases where this call would fail but native "git" call does not.
             new DescribeCommand(git.getRepository()).call()
 
-            String version = runGitCommand(project.rootDir, "describe", "--tags", "--always", "--first-parent") ?: UNSPECIFIED_VERSION
-            boolean isClean = git.status().call().isClean()
-            return version + (isClean ? '' : '.dirty')
+            return GitCli.runGitCommand(project.rootDir, "describe", "--tags", "--always", "--first-parent")
         } catch (Throwable t) {
-            return UNSPECIFIED_VERSION
+            return null
```
uschi2000, line 79: not a big fan of the nulls everywhere. isn't reasonable to just fail instead of returning nulls (both here and below)?

uschi2000 (changes requested):

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
iamdanfox: I'm matching behaviour here - when these commands fail, we output the string 'unspecified'.

uschi2000:
OK

iamdanfox:
@uschi2000 if you're happy with this, can you hit merge? I don't have any permissions on this repo.

uschi2000 (approved):

Issue #45, opened 2017-03-05
Version is Wrong when non-version Tags exist


pluttrell:
This plugin works great when only version specific tags exist. If there are other tags in the repo, then it doesn't work properly, for example it would created `featureX-5-g93f233e1` if someone tagged a commit with `featureX`. 

Perhaps this plugin could define (or better yet, allow users to define) a prefix for which tags it considers to be tags with versions in them. For example if `v` were used, then a valid version tag would be `v0.1`.

jchamberlain:
Unfortunately, the merged support for prefixes does not fix this problem for anyone using standard version tags in the format `X.Y.Z` or `vX.Y.Z`. `X.Y.Z` doesn't have a prefix, so you can't limit to tags of that format. `v` is not an allowed prefix, so you're in the same boat.

I see #61 took a stand against the `v` prefix, but is that set in stone?

PR #61, opened 2017-06-24
Allow definition of a tag prefix

MatthewMaclean:
gitVersionOpt is added which allows for configuration of prefix and
includePrefix. By specifying a prefix, only tags including that prefix
will be searched. If includePrefix is by default false, which means that
it will be stripped from the output version. e.g. v0.10.0 -> 0.10.0

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -39,6 +39,14 @@ details.branchName // is null if the repository in detached HEAD mode
 details.isCleanTag
 ```
 
+You can optionally search a subset of tags with `prefix`. Inclusion of the `prefix` in the returned version is toggled
+with `includePrefix` (default: `false`). Example when the tag is my-product2.15.0:
+
+```groovy
+gitVersion(prefix:'my-product') // -> 2.15.0
+gitVersion(prefix:'my-product', includePrefix:true) // -> my-product2.15.0
+```
+
 Tasks
 -----
 This plugin adds a `printVersion` task, which will echo the project's configured version
--- a/readme.md
+++ b/readme.md
@@ -39,10 +39,18 @@ details.branchName // is null if the repository in detached HEAD mode
 details.isCleanTag
 ```
 
-You can optionally search a subset of tags with `prefix`. Example when the tag is my-product2.15.0:
+You can optionally search a subset of tags with `prefix`. Example when the tag is my-product@2.15.0:
 
 ```groovy
-gitVersion(prefix:'my-product') // -> 2.15.0
+gitVersion(prefix:'my-product@') // -> 2.15.0
+```
+
+Valid prefixes are defined by the regex `[/@]?([A-Za-z]+[/@-])+`.
+```
+/Abc/
+Abc@
+foo-bar@
+foo/bar@
 ```
 
 Tasks
--- a/readme.md
+++ b/readme.md
@@ -39,12 +39,10 @@ details.branchName // is null if the repository in detached HEAD mode
 details.isCleanTag
 ```
 
-You can optionally search a subset of tags with `prefix`. Inclusion of the `prefix` in the returned version is toggled
-with `includePrefix` (default: `false`). Example when the tag is my-product2.15.0:
+You can optionally search a subset of tags with `prefix`. Example when the tag is my-product2.15.0:
 
 ```groovy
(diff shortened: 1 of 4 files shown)


MatthewMaclean:
Relates to: https://github.com/palantir/gradle-git-version/issues/45
Also useful when using https://github.palantir.build/acme/lerna which creates tags for each package.

MatthewMaclean:
Relates to: #45
Also useful when using https://github.com/lerna/lerna which creates tags for each package.

readme.md
```diff
@@ -39,6 +39,16 @@ details.branchName // is null if the repository in detached HEAD mode
 details.isCleanTag
 ```
 
+You can optionally configure the plugin to search a subset of tags with `prefix`. Inclusion of the `prefix` in the returned version is toggled
+with `includePrefix` (default: `false`). Example configuration:
+
+```groovy
+gitVersionOpt {
+    prefix = "v"
```
uschi2000: I don't really want folks to tag vX.Y.Z. Why do we need this?

MatthewMaclean: I can switch the example to what Lerna does, where it will create a tag for each package in the repo. So I could have:

Product@X.Y.Z
product-api@A.B.C
product-app@C.D.E

This allows you to easily navigate to the commit where a version was bumped for any package in the repository. This means that gitVersion() won't work because it assumes you are only tracking versions for one thing in the repository.

uschi2000:
OK. Isn't the right API for this something like:

project.version = gitVersion(prefix='my-product')

?

MatthewMaclean:
Yep, didn't know about named parameters but they're way nicer. Added some tests and updated documentation.

readme.md
```diff
@@ -39,6 +39,14 @@ details.branchName // is null if the repository in detached HEAD mode
 details.isCleanTag
 ```
 
+You can optionally search a subset of tags with `prefix`. Inclusion of the `prefix` in the returned version is toggled
+with `includePrefix` (default: `false`). Example when the tag is my-product2.15.0:
+
+```groovy
+gitVersion(prefix:'my-product') // -> 2.15.0
+gitVersion(prefix:'my-product', includePrefix:true) // -> my-product2.15.0
```
uschi2000: i'd prefer to be a little more opinionated/constraining here. in particular, i don't want projects to tag `v3.1.0` in order to produce version `3.1.0` with `gitVersion(prefix:'v')`. further, do we need to support both keeping-the-prefix and stripping-the-prefix modes?

proposal: force prefixes to match exactly `[a-z]+-`.

MatthewMaclean: For supporting both keeping the prefix and stripping it, I'm fine with just stripping it. Since you are grabbing the version for a certain product based on the prefix, returning just the version is acceptable to me.

--------------

What's the reason for so strongly trying to avoid the `v` prefix in git tags if it won't affect the published version? 

An example Lerna tag is: `@npmorg/product-api@X.Y.Z`

I think it's possible to focus around a delimiter between the prefix and the version, so:
`.+[-\.@_]$`. This would allow the prefix to be whatever you want, but it avoids the v3.1.0 case. We may run into the occasional PR adding a delimiter to the regex.

uschi2000: Rationale: the tooling should guide users towards standard best-practices. For us, they are to use X.Y.Z, not vX.Y.Z.

Let's also constrain the prefix a bit more, how about this: [a-z]+([-/][a-z]+)*@
Good:
```
abc@
abc@
foo-bar@
foo/bar@
```

Bad
```
foo-
foo@bar@
```

uschi2000 (approved):

readme.md
john-tipper: @uschi2000 What's the reasoning behind the prefix regex of `[/@]?([A-Za-z]+[/@-])+` please?
If we mandate a delimiter after the prefix (i.e. `[/@-]`), then why should we not allow a numeric value?

Use case: I have 2 modules in a Gradle build (`mod`, `mod2`) and I want a prefix to distinguish between the 2 modules.  I can't change the module names, that's not up to me; btw, `mod2` is a newer/better version of `mod`.

Why could the regex not be `[/@]?([A-Za-z0-9]+[/@-])+` ?

uschi2000: I think we can allow numbers.

PR #68, opened 2017-09-01
Add non-abbreviated GitHash as gitHashFull

### LINKED ISSUE #67: GitHash should allow non-abbreviation
Right now gitHash is hard-coded to return a 10 character abbreviation. But this should be allowed to be configured with a default.
This is useful when we want to convert it back to an ObjectID
https://github.com/palantir/gradle-git-version/blob/develop/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy#L104

matthew-dailey:
* Adds `versionDetails().gitHashFull` to return the 40-character git hash.  This is a backwards-compatible change.
* Adds `gitFullHash` to constructor for `VersionDetails`.  _If_ `VersionDetails` is part of the public API, then this would be a backwards-_incompatible_ change
* Did _not_ update the README because this feature would need to be part of a release before it is available to users
* Closes #67

### DIFF
--- a/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
@@ -64,10 +64,11 @@ class GitVersionPlugin implements Plugin<Project> {
         verifyPrefix(args.prefix)
         String description = stripPrefix(gitDescribe(project, args.prefix), args.prefix)
         String hash = gitHash(project)
+        String fullHash = gitHashFull(project)
         String branchName = gitBranchName(project)
         boolean isClean = isClean(project)
 
-        return new VersionDetails(description, hash, branchName, isClean)
+        return new VersionDetails(description, hash, fullHash, branchName, isClean)
     }
 
     @Memoized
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-git-version](https://github.com/palantir/gradle-git-version), @matthew-dailey! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -105,6 +106,16 @@ class GitVersionPlugin implements Plugin<Project> {
     }
 
     @Memoized
+    private String gitHashFull(Project project) {
```
uschi2000: can you change `gitHash()` to use this function, please?

uschi2000 (approved):

basil:
Could we please get a release with this feature available? I'd like to use it on a project I'm working on, but the latest release doesn't have this commit.

uschi2000:
Done, 0.9.0

basil:
I just upgraded, and it works. Thanks a lot!

PR #74, opened 2017-11-08
Allow forcing snapshot version

dansanduleac:
When you pass `forceSnapshot: true` to `versionDetails` or `gitVersion`, if HEAD points to a tag (or more), exclude these tags from git-describe's algorithm.

### DIFF
--- a/src/main/groovy/com/palantir/gradle/gitversion/GitCli.groovy
+++ b/src/main/groovy/com/palantir/gradle/gitversion/GitCli.groovy
@@ -15,8 +15,14 @@
  */
 package com.palantir.gradle.gitversion
 
+import org.gradle.api.Project
+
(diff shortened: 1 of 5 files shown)


uschi2000:
so this gives me the snapshot version since the last tag? why would i need/want this?

dansanduleac:
In order to ensure that, after you tag a branch's head and push both, the branch build in circle can't pick up the tag and try to publish it, if the branch's build starts more or less at the same time as the tag's build.
Sadly it doesn't seem to work with git 1.x :(

uschi2000:
I’d prefer solving this by running the publish step on tags only in Circle config.

uschi2000:
discussed offline, no need to do this any more.

PR #77, opened 2017-11-19
Use rev-list --first-parent to support git < 1.8.4

abless:
Current implementation uses --first-parent for git describe, a feature only added in Git 1.8.4. This change uses rev-list instead to walk the commit history and format the version string.

Tested with Git versions 1.7.1 and 2.11.1

### DIFF
--- a/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
+++ b/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
@@ -15,6 +15,7 @@
  */
 package com.palantir.gradle.gitversion
 
+import com.google.common.base.Splitter
 import groovy.transform.Memoized
 import org.eclipse.jgit.api.DescribeCommand
 import org.eclipse.jgit.api.Git
@@ -25,8 +26,11 @@ import org.eclipse.jgit.lib.Ref
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 
+import java.util.stream.Collectors
+
 class GitVersionPlugin implements Plugin<Project> {
 
+    private static final int SHA_ABBR_LENGTH = 7
     private static final int VERSION_ABBR_LENGTH = 10
     private static final String PREFIX_REGEX = "[/@]?([A-Za-z]+[/@-])+"
 
@@ -79,12 +83,6 @@ class GitVersionPlugin implements Plugin<Project> {
 
     @Memoized
     private String gitDescribe(Project project, String prefix) {
-        def abbrevHash = { it.substring(0, 7) }
-
-        def formatDescription = { tag, depth, commitHash ->
-            depth == 0 ? tag : String.format("%s-%s-g%s", tag, depth, abbrevHash(commitHash))
-        }
-
         // verify that "git" command exists (throws exception if it does not)
         GitCli.verifyGitCommandExists()
 
@@ -98,15 +96,24 @@ class GitVersionPlugin implements Plugin<Project> {
              * Mimick 'git describe --tags --always --first-parent --match=${prefix}*' by using rev-list to
              * support versions of git < 1.8.4
              */
-            int depth = 0
-            String revList = GitCli.runGitCommand(project.rootDir, "rev-list", "--first-parent", "HEAD")
-            for (String rev : revList.split(System.getProperty("line.separator"))) {
-                String exactTag = GitCli.runGitCommand(project.rootDir, "describe", "--tags", "--exact-match",
-                        "--match=${prefix}*", rev)
-                if (exactTag != "") {
-                    return formatDescription(exactTag, depth, rev)
+
+            // Get SHAs of all tags, we only need to search for these later on
+            List<String> tagList = getLines(GitCli.runGitCommand(project.rootDir, "show-ref", "--tags", "-d"))
+            Set<String> tagRefs = tagList.stream()
+                    .map{Splitter.on(' ').splitToList(it)}
+                    .filter{it.size() == 2}
+                    .filter{it.get(1).matches("^refs/tags/${prefix}.*")}
+                    .map{it.get(0) }
+                    .collect(Collectors.toSet())
+
+            List<String> revs = getLines(GitCli.runGitCommand(project.rootDir, "rev-list", "--first-parent", "HEAD"))
+            for (int depth = 0; depth < revs.size(); depth++) {
+                String rev = revs.get(depth)
+                if (tagRefs.contains(rev)) {
+                    String exactTag = GitCli.runGitCommand(project.rootDir, "describe", "--tags", "--exact-match",
+                            "--match=${prefix}*", rev)
+                    return depth == 0 ? exactTag : String.format("%s-%s-g%s", exactTag, depth, abbrevHash(rev))
                 }
-                depth += 1
             }
 
             // No tags found, so return commit hash of HEAD
@@ -114,6 +121,16 @@ class GitVersionPlugin implements Plugin<Project> {
         } catch (Throwable t) {
             return null
         }
+    }
+
+    @Memoized
+    private List<String> getLines(String s) {
+        return Splitter.on(System.getProperty("line.separator")).splitToList(s)
+    }
+
+    @Memoized
+    private String abbrevHash(String s) {
+        return s.substring(0, SHA_ABBR_LENGTH)
     }
 
     @Memoized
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -79,6 +79,12 @@ class GitVersionPlugin implements Plugin<Project> {
 
     @Memoized
     private String gitDescribe(Project project, String prefix) {
+        def abbrevHash = { it.substring(0, 7) }
```
justinuang: why not just make this a method?

justinuang: Pull out 7 as a constant.

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -79,6 +79,12 @@ class GitVersionPlugin implements Plugin<Project> {
...
+
+        def formatDescription = { tag, depth, commitHash ->
```
justinuang: Also pull this out into a method?

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -88,8 +94,23 @@ class GitVersionPlugin implements Plugin<Project> {
...
             new DescribeCommand(git.getRepository()).call()
 
-            return GitCli.runGitCommand(project.rootDir, "describe", "--tags", "--always", "--first-parent",
-                    "--match=${prefix}*")
+            /*
+             * Mimick 'git describe --tags --always --first-parent --match=${prefix}*' by using rev-list to
+             * support versions of git < 1.8.4
+             */
+            int depth = 0
+            String revList = GitCli.runGitCommand(project.rootDir, "rev-list", "--first-parent", "HEAD")
```
iamdanfox: Tiny concern that this might be a tiny bit slower by loading up all commits on develop, but I tried this on a 2 year old internal repo and it was only 2000ish hashes, so probably nbd.

If it turns out to be a problem I'm sure we can page our way through the rev-list

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -88,8 +94,23 @@ class GitVersionPlugin implements Plugin<Project> {
...
+            for (String rev : revList.split(System.getProperty("line.separator"))) {
+                String exactTag = GitCli.runGitCommand(project.rootDir, "describe", "--tags", "--exact-match",
+                        "--match=${prefix}*", rev)
+                if (exactTag != "") {
+                    return formatDescription(exactTag, depth, rev)
```
iamdanfox: Would prefer to inline this because it's such a short function and the lambda is not used anywhere else.

iamdanfox:
I'd be a bit happier merging this if we could run the tests against an old version of git on CI somehow... otherwise I think it's likely that someone will unintentionally use a newer git feature in a future PR and thereby break the compatibility you need.

We could either use a docker-container for this, or maybe change the installed version of git on CircleCI.

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -88,8 +94,23 @@ class GitVersionPlugin implements Plugin<Project> {
...
 
-            return GitCli.runGitCommand(project.rootDir, "describe", "--tags", "--always", "--first-parent",
-                    "--match=${prefix}*")
+            /*
+             * Mimick 'git describe --tags --always --first-parent --match=${prefix}*' by using rev-list to
+             * support versions of git < 1.8.4
+             */
+            int depth = 0
+            String revList = GitCli.runGitCommand(project.rootDir, "rev-list", "--first-parent", "HEAD")
+            for (String rev : revList.split(System.getProperty("line.separator"))) {
```
justinuang: Use guava Splitter

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -88,8 +94,23 @@ class GitVersionPlugin implements Plugin<Project> {
...
+                String exactTag = GitCli.runGitCommand(project.rootDir, "describe", "--tags", "--exact-match",
```
justinuang: I'm a little worried this is slow.

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
iamdanfox: Rather than a separate `int depth = 0` declaration and this nice java for each loop, what about using an old school `for(int depth=0; depth < revListArray.length; depth++)` to emphasise that depth is just the index of the hash that we're looking at?

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -88,8 +94,23 @@ class GitVersionPlugin implements Plugin<Project> {
...
+                        "--match=${prefix}*", rev)
+                if (exactTag != "") {
+                    return formatDescription(exactTag, depth, rev)
+                }
+                depth += 1
```
justinuang: We need to handle dirty commits as well

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
justinuang: Do we have metrics on how long this will take?

justinuang (changes requested):

abless:
@justinuang is right, this becomes inefficient if the latest tag has a long commit distance since we're spawning a new process every time. I tested on a repo with 1000 commits (and no tags), and it takes ~6s to produce the version string.

Here's an alternative proposal that should be faster: 

1. Still call rev-list to get list of all commits. As @iamdanfox pointed out we might want to add paging at some point, but probably overkill for now.
2. Call git show-ref --tags to get the list of all tags with their commit hash.
3. Find the most recent commit in rev-list that also is a tag and format appropriately.

It's a bit silly that JGit doesn't seem to support this natively.

abless:
This is actually more involved since we might have to deal with both annotated and un-annotated tags and also pick the latest tag by creation date. I think it's possible to get all that information using git for-each-ref and formatting appropriately, but it's starting to look nasty...

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
abless: That's already handled in VersionDetails (we also have a check for that)

abless:
Never mind, I realized we can still rely on "git describe" to pick the right tag for us if there are multiple. I just added an additional step that allows us to only call "git describe" once.

I tested this new logic on a repo with a tag followed by ~8000 commits and it took about ~600ms to describe the version (for reference, git describe in the command line took ~360ms).

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -88,13 +91,47 @@ class GitVersionPlugin implements Plugin<Project> {
...
-                    "--match=${prefix}*")
+            /*
+             * Mimick 'git describe --tags --always --first-parent --match=${prefix}*' by using rev-list to
+             * support versions of git < 1.8.4
+             */
+
+            // Get SHAs of all tags, we only need to search for these later on
+            Set<String> tagRefs = Sets.newHashSet()
+            for (String tag : getLines(GitCli.runGitCommand(project.rootDir, "show-ref", "--tags", "-d"))) {
+                List<String> parts = Splitter.on(' ').splitToList(tag)
```
justinuang: pull out Splitter.on() into a static var, same in getLines as well. Perf optimizations =)

justinuang: https://github.com/google/guava/wiki/StringsExplained

justinuang (changes requested):

justinuang:
lgtm, @iamdanfox can you finish the review and merge if it looks good/

iamdanfox (approved):
Code looks solid :) I'd still feel better if we had a test that runs this against an old version of git, but if you're OK with this going straight to your prod @abless then I'm happy to tag a release! (Clearly all the existing tests still pass!)

PR #81, opened 2018-01-09
Fix wrong commit sha in version

abless:
Current implementation has a bug, in that the abbreviated hash is that of the latest tag instead of the latest commit. This fixes the issue and also includes a test.

### DIFF
--- a/src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
@@ -160,7 +160,7 @@ class GitVersionPluginTests extends Specification {
         buildResult.output.contains(":printVersion\n1.0.0\n")
     }
 
-    def 'git describe when unannotated tag is present' () {
+    def 'git describe when lightweight tag is present' () {
         given:
         buildFile << '''
             plugins {
@@ -489,7 +489,7 @@ class GitVersionPluginTests extends Specification {
         buildResult.output.contains(":printVersion\n1.0.0-1-g${commitSha.substring(0, 7)}\n")
     }
 
-    def 'git describe with commit after unannotated tag' () {
+    def 'git describe with commit after lightweight tag' () {
         given:
         buildFile << '''
             plugins {
--- a/src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
+++ b/src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
@@ -160,6 +160,27 @@ class GitVersionPluginTests extends Specification {
         buildResult.output.contains(":printVersion\n1.0.0\n")
     }
 
+    def 'git describe when unannotated tag is present' () {
+        given:
+        buildFile << '''
+            plugins {
+                id 'com.palantir.git-version'
+            }
+            version gitVersion()
+        '''.stripIndent()
+        gitIgnoreFile << 'build'
+        Git git = Git.init().setDirectory(projectDir).call();
+        git.add().addFilepattern('.').call()
+        git.commit().setMessage('initial commit').call()
+        git.tag().setAnnotated(false).setName('1.0.0').call()
+
+        when:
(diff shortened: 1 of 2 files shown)


abless:
@iamdanfox for approval/merge

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -442,6 +442,31 @@ class GitVersionPluginTests extends Specification {
...
+        git.add().addFilepattern('.').call()
+        git.commit().setMessage('initial commit').call()
+        git.tag().setAnnotated(true).setMessage('1.0.0').setName('1.0.0').call()
+        dirtyContentFile << 'dirty-content'
+        git.add().addFilepattern('.').call()
+        git.commit().setMessage('added some stuff').call()
+
+        when:
+        BuildResult buildResult = with('printVersion').build()
+        String commitSha = (++git.log().call().iterator()).getName()
```
iamdanfox: I haven't seen this '++' syntax used this way before, is there any way to get the most recent commit SHA out of the `git.commit()` invocation above instead?

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -442,6 +442,31 @@ class GitVersionPluginTests extends Specification {
         buildResult.output =~ ":printVersionDetails\n1.0.0\n"
     }
 
+    def 'git describe with commit after tag' () {
```
iamdanfox: :+1: for the integ test!!

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -442,6 +442,31 @@ class GitVersionPluginTests extends Specification {
...
+            plugins {
+                id 'com.palantir.git-version'
+            }
+            version gitVersion()
+        '''.stripIndent()
+        gitIgnoreFile << 'build'
+        Git git = Git.init().setDirectory(projectDir).call();
+        git.add().addFilepattern('.').call()
+        git.commit().setMessage('initial commit').call()
+        git.tag().setAnnotated(true).setMessage('1.0.0').setName('1.0.0').call()
```
iamdanfox, line 479: Is it worth verifying the code works for a non annotated tag too?

abless: Is that not tested already? If not, we should probably add it but it'd be independent of this change IMHO.

iamdanfox: Given that this PR touches the `%s-%s-g%s` code, I think it's worth proving that it works correctly when it encounters both annotated and non-annotated tags.  I couldn't see an existing test case that captures this - all the others just do `git.tag().setAnnotated(true)`.

If it turns out to require other code-changes to pass, then I'm happy to put that test in a separate PR.

abless: Ok sure, I'll add tests for both cases (the exact tag case and the %s-%s-g%s case).

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -159,6 +160,27 @@ class GitVersionPluginTests extends Specification {
         buildResult.output.contains(":printVersion\n1.0.0\n")
     }
 
+    def 'git describe when unannotated tag is present' () {
```
iamdanfox: I just realised why 'non-annotated' tag sounds weird... apparently they're called '[lightweight tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging#_lightweight_tags)'

iamdanfox (approved):
Looks great, thanks for the comprehensive tests!

PR #86, opened 2018-04-17
Native vs Jgit

### LINKED ISSUE #85: Fallback to jgit if git command doesn't exist
Any objections to adding fallback to jgit if `GitCli.verifyGitCommandExists()` fails [here](https://github.com/palantir/gradle-git-version/blob/develop/src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy#L87)?
@iamdanfox @uschi2000

delta003:
Fix #85


palantirtech:
Thanks for your interest in [palantir/gradle-git-version](https://github.com/palantir/gradle-git-version), @delta003! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

delta003:
So tests are passing with jgit. I can walk you through it tmrw and we can decide how to move forward.

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -79,65 +72,14 @@ class GitVersionPlugin implements Plugin<Project> {
...
-
-    @Memoized
-    private List<String> getLines(String s) {
-        return LINE_SPLITTER.splitToList(s)
-    }
-
-    @Memoized
-    private String abbrevHash(String s) {
-        return s.substring(0, SHA_ABBR_LENGTH)
+        // return new NativeGitDescribe(project.projectDir).describe(prefix)
```
helenyuyu: did you mean to keep this comment?

delta003: It's really scary to break this plugin considering it's used everywhere, and I'm not expert on git/jgit to be sure I've handled all edge cases.
@iamdanfox and I were thinking of some safe approach:
- compute both
- if one returns null, use the other one
- if both commands succeed, check if output is same
- if it's not, raise an exception (so we can fix the jgit implementation)
- eventually drop native one, once we're confident we have the same behaviour

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
```diff
@@ -0,0 +1,81 @@
...
+        }
+        RevWalk walk = new RevWalk(git.getRepository())
+        RevCommit commit = walk.parseCommit(headObjectId)
+        List<String> revs = new ArrayList<>()
+        while (commit) {
+            revs.add(commit.getName())
+            try {
+                commit = commit.getParent(0)
+            } catch (Throwable ignored) {
+                commit = null
```
iamdanfox: break

delta003:
@iamdanfox `git describe` has defined behaviour when it comes to multiple tags on same commit. I wrote a comparator that does the same. Added some tests to cover that

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -79,67 +72,28 @@ class GitVersionPlugin implements Plugin<Project> {
...
-                tagRefs.add(parts.get(0))
+        // TODO(mbakovic): Switch to jgit only implementation
+        String nativeGitDescribe = new NativeGitDescribe(project.projectDir).describe(prefix)
+        String jgitDescribe = new JGitDescribe(project.projectDir).describe(prefix)
+        if (nativeGitDescribe == null) {
+            return jgitDescribe
+        } else if (jgitDescribe == null) {
+            return nativeGitDescribe
+        } else {
+            if (nativeGitDescribe != jgitDescribe) {
```
iamdanfox: use `.equals` here to compare strings instead of `!=`

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -79,67 +72,28 @@ class GitVersionPlugin implements Plugin<Project> {
...
+                throw new IllegalStateException(String.format(
+                        "Inconsistent git describe, native was %s and jgit was %s", nativeGitDescribe, jgitDescribe))
```
iamdanfox: Thoughts on adding "Please report this on `github.com/palantir/gradle-git-version`"

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
```diff
@@ -0,0 +1,101 @@
...
+    private File directory
+
+    JGitDescribe(File directory) {
+        this.directory = directory
+    }
+
+    @Override
+    String describe(String prefix) {
+        Git git = Git.wrap(new FileRepository(GitCli.getRootGitDir(directory)))
+        if (!GitUtils.isBackCompatible(git)) {
```
iamdanfox: Not a huge fan of the method name here... why not just call this `isRepoEmpty` ? I don't think the log line should mention back compatibility - users don't really care what the old versions did.

src/main/groovy/com/palantir/gradle/gitversion/GitDescribe.groovy
```diff
@@ -0,0 +1,6 @@
+package com.palantir.gradle.gitversion
+
+interface GitDescribe {
+
+    String describe(String prefix)
```
iamdanfox: I think it would be helpful to add some javadoc explaining that this is effectively `git describe --tags --always --first-parent --match=${prefix}*`, and should return null for empty repos.

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
```diff
@@ -0,0 +1,101 @@
...
+import org.eclipse.jgit.lib.ObjectId
+import org.eclipse.jgit.lib.PersonIdent
+import org.eclipse.jgit.lib.Ref
+import org.eclipse.jgit.revwalk.RevCommit
+import org.eclipse.jgit.revwalk.RevTag
+import org.eclipse.jgit.revwalk.RevWalk
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class JGitDescribe implements GitDescribe {
```
iamdanfox: I think we need to give future readers some javadoc to help them understand why we built this - specifically that shelling out to `git` is a bit unsafe because older versions of git don't support `--first-parent` and jgit doesn't support this natively (https://github.com/palantir/gradle-git-version/issues/46)

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
```diff
@@ -0,0 +1,101 @@
...
+        ObjectId headObjectId
+        try {
+            headObjectId = git.getRepository().resolve(Constants.HEAD)
+        } catch (Throwable ignored) {
+            log.debug("HEAD not found")
+            return null
+        }
+
+        try {
+            RevWalk walk = new RevWalk(git.getRepository())
```
iamdanfox: Could we extract these little blocks into methods, e.g. `revList`, `computeTags` ?

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -573,11 +574,95 @@ class GitVersionPluginTests extends Specification {
...
+        git.tag().setAnnotated(false).setName('1.0.0').call()
+        git.tag().setAnnotated(false).setName('3.0.0').call()
+
+        when:
+        BuildResult buildResult = with('printVersion').build()
+
+        then:
+        buildResult.output.contains(":printVersion\n1.0.0\n")
+    }
+
```
iamdanfox, line 658: Could we add a test where the git tree contains a symlink to another file? Apparently that was historically an incompatibility between real git and jgit

src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
delta003: +1

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
delta003: Makes sense. Let's not log anything

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
delta003: Yes. I'll also add comment in GitVersionPlugin.gitDescribe to explain running both native and jgit

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
delta003: Looks much better

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
delta003: turns out there is already a test for that https://github.com/palantir/gradle-git-version/blob/develop/src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy#L277

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
```diff
@@ -0,0 +1,121 @@
...
+        Git git = Git.wrap(new FileRepository(GitCli.getRootGitDir(directory)))
+        if (!GitUtils.isRepoEmpty(git)) {
+            log.debug("Repository is empty")
+            return null
+        }
+
+        RevCommit headCommit
+        try {
+            ObjectId headObjectId = git.getRepository().resolve(Constants.HEAD)
+            RevWalk walk = new RevWalk(git.getRepository())
```
delta003: @iamdanfox there is some groovy weirdness happening here because walk is used in line 47

iamdanfox: Yeah I just noticed that, will revert my commit but I'm kinda horrified the tests are passing

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
```diff
@@ -0,0 +1,123 @@
...
+            }
+        }
+        return revs
+    }
+
+    // Maps all commits returned by 'git show-ref --tags -d' to output of 'git describe --tags --exact-match <commit>'
+    private Map<String, String> mapCommitsToTags(Git git, RevWalk walk) {
+        // Maps commit hash to list of all refs pointing to given commit hash.
+        // All keys in this map should be same as commit hashes in 'git show-ref --tags -d'
+        Map<String, List<RefWithTagName>> commitHashToTags = new HashMap<>()
```
iamdanfox, line 90: It seems a tiny bit unnecessary to build up a whole list of `RefWithTagNames` and then use the comparator to pull out the minimum - in theory we could just do the comparison whenever we get a new tag so that we only ever hold on to one.

Not sure it's worth blocking the PR for this - I think I'd like to convert everything to java before we try to do any more refactoring!

iamdanfox (approved):
Let's ship an RC and find out!

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
delta003: There is a huge bug. Tests aren't running correctly :/ Do not merge

src/main/groovy/com/palantir/gradle/gitversion/JGitDescribe.groovy
delta003: I really want to rewrite this in Java 8... But addressed this in #89

PR #89, opened 2018-04-19
Native vs JGit 2

delta003:
There was a bug in https://github.com/palantir/gradle-git-version/pull/86 so it's reverted


src/main/groovy/com/palantir/gradle/gitversion/GitVersionPlugin.groovy
```diff
@@ -79,65 +72,35 @@ class GitVersionPlugin implements Plugin<Project> {
...
-                    }
-                }
-            }
-
-            // No tags found, so return commit hash of HEAD
-            return abbrevHash(runGitCmd("rev-parse", "HEAD"))
-        } catch (Throwable t) {
-            return null
+        // If native succeeded, make sure it's same as JGit one
+        if (!nativeGitDescribe.equals(jgitDescribe)) {
```
delta003, line 96: This is actually behaviour we want. If native succeeds, it must be equal with jgit. So if jgit fails and native succeeds we'll raise an exception. That should never happen!

delta003:
Also replace all Throwable with Exception...

delta003:
Now it's ready

iamdanfox (approved):

PR #90, opened 2018-04-20
Switch to java

### LINKED ISSUE #70: Switch plugin source to Java
Gradle performance guide recommends using Java for plugins to avoid a small Groovy overhead:

> However, it’s important to bear in mind that there is a small cost associated with method calls in dynamic Groovy.

https://guides.gradle.org/performance/#statically_compile_tasks_and_plugins

iamdanfox:
Gradle themselves recommend writing plugins in Java.  You get a slight performance boost and easier refactoring.  Tests are still in groovy.

No logic changes, just lots of semicolons.

fixes https://github.com/palantir/gradle-git-version/issues/70


delta003 (approved):
Much better!

iamdanfox:
This loses the magic groovy `@memoized` stuff so as a followup I'd like to move all that logic into one immutables generated class and use `@lazy` to get the same behaviour.  This should also make the top level plugin much clearer!

PR #93, opened 2018-04-20
Unit tests & reorganize classes

iamdanfox:
Previously, logic was scattered across a number of different classes: the top level plugin, two utility classes and VersionDetails.

Now, logic is consolidated as much as possible.  VersionDetails computes all the information that is returned to users and utilities have been removed where possible.

This allows us to run some unit tests from IDEs


delta003 (approved):
lgtm

PR #94, opened 2018-04-24
Use only jgit

### LINKED ISSUE #87: Use JGit implementation only
See `GitVersionPlugin.gitDescribe`

delta003:
Fix #87


src/main/java/com/palantir/gradle/gitversion/JGitDescribe.java
```diff
@@ -64,6 +64,7 @@ public String describe(String prefix) {
 
         Repository repo = git.getRepository();
         try (RevWalk walk = new RevWalk(repo)) {
+            walk.setRetainBody(false);
```
delta003, line 67: Just an optimization

src/main/java/com/palantir/gradle/gitversion/JGitDescribe.java
```diff
@@ -76,6 +77,8 @@ public String describe(String prefix) {
 
                 head = walk.parseCommit(parents[0]);
             }
+
+            walk.close();
```
delta003: I don't think this is required but it's safe to close this walk.

src/main/java/com/palantir/gradle/gitversion/JGitDescribe.java
```diff
@@ -88,16 +91,19 @@ public String describe(String prefix) {
...
-        for (Map.Entry<String, Ref> entry : git.getRepository().getTags().entrySet()) {
-            RefWithTagName refWithTagName = new RefWithTagName(entry.getValue(), entry.getKey());
-
-            ObjectId peeledRef = refWithTagName.getRef().getPeeledObjectId();
-            if (peeledRef == null) {
-                // lightweight tag (commit object)
-                updateCommitHashMap(commitHashToTag, comparator, entry.getValue().getObjectId(), refWithTagName);
+        Repository repository = git.getRepository();
+        for (Map.Entry<String, Ref> entry : repository.getTags().entrySet()) {
+            Ref peeledRef = repository.peel(entry.getValue());
```
delta003, line 94: Found a bug while writing unit tests. This is done for us when we run `git.describe().call()`. But we shouldn't depend on that. Peeling ref means following links and actually getting peeled object id for this ref. Otherwise, `peeledRef.getPeeledObjectId()` would just return null

src/main/java/com/palantir/gradle/gitversion/VersionDetails.java
```diff
@@ -141,4 +125,19 @@ public String getBranchName() throws IOException {
 
         return ref.getName().substring(Constants.R_HEADS.length());
     }
+
+    @Override
+    public String toString() {
```
delta003, line 130: This object used to have nice representation when it was written in Groovy. Trying to match that representation with this toString

src/main/java/com/palantir/gradle/gitversion/JGitDescribe.java
iamdanfox: this is definitely not necessary, the try-with-resources block will call `walk.close()` automatically.

src/main/java/com/palantir/gradle/gitversion/JGitDescribe.java
iamdanfox: 😱

src/test/java/com/palantir/gradle/gitversion/JGitDescribeTest.java
```diff
@@ -0,0 +1,223 @@
...
+    private PersonIdent identity = new PersonIdent("name", "email@address");
+
+    @Before
+    public void before() throws GitAPIException {
+        projectDir = temporaryFolder.getRoot();
+        git = Git.init().setDirectory(projectDir).call();
+    }
+
+    @Test
+    public void test_on_annotated_tag() throws Exception {
```
iamdanfox, line 41: Should we also be deleting the corresponding tests from the gradle integration tests?

iamdanfox: Or do we want to leave the test coverage there?  (e.g. symlinks and dirty tree?)

delta003: I kind of see them more as integration tests. They actually apply plugin and execute gradle tasks so I'd keep them.

iamdanfox (approved):

PR #95, opened 2018-04-30
Upgrade jgit

iamdanfox:
Note this isn't latest jgit because it requires a java8 compiler.

Apparently some bugs were fixed: https://bugs.eclipse.org/bugs/buglist.cgi?classification=Technology&list_id=10006180&order=Importance&product=JGit&query_format=advanced&resolution=FIXED&resolution=DUPLICATE&target_milestone=4.5

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ repositories {
 
 dependencies {
     compile gradleApi()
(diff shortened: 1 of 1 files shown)


delta003 (approved):
We should consider migrating to java8 with 1.0

iamdanfox:
yeah I think that should be doable as much of our other gradle tooling requires Java8.

PR #96, opened 2018-04-30
Re-create a `Git` object every time tasks are invoked

iamdanfox:
Suspicious report on 0.11.0-rc6:

```
dirty file-tree
runs git clean -fd
versionDetails.isCleanTag returns false
```

Current working guess is that JGit might be caching something


delta003 (approved):
Btw this shouldn't be too hard to repro in tests. But I think not caching anything is correct behaviour for these tasks. Each task must be executed from scratch every time as things might have changed.

PR #103, opened 2018-08-29
move tag-prefix filtering to `mapCommitsToTags`

### LINKED ISSUE #34: version could be wrong when there are more than 1 tags on the commit
when there's `1.9.0` and `1.10.0` as annotated tags for a commit, `1.9.0` will be returned (probably because it's the first thing in a list thats sorted alphabetically). it should probably be sorted via semver and always return the larger version instead. 

notice that this only happens because of the string "9" is greater than "1". it's fine on other versions

philgebhardt:
allowing proper filtering with multiple tags per commit

fixes https://github.com/palantir/gradle-git-version/issues/34


palantirtech:
Thanks for your interest in [palantir/gradle-git-version](https://github.com/palantir/gradle-git-version), @philgebhardt! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

martin-a:
Could someone look into this PR ? #34 is causing us some troubles as well.

philgebhardt:
Closed because I've decided I'd rather not contribute to Palantir

PR #128, opened 2019-02-20
[improvement] Cache result of git calls

### LINKED ISSUE #106: Cache gitVersion() result
if you do

```
allprojects {
  version gitVersion()
}
```

you will compute version for every subproject even though they're in the same repo. This increases the runtime by around ~0.5s per subproject. We should cache the result since it won't change across projects.

raiju:
Fixes #106

## Before this PR

Git calls aren't cached.

## After this PR

Git call results are cached, improving build speeds in multi-project setups even when using the plugin naively.

The cache is both version arg & git root aware, so will handle e.g. different prefixes/git roots being used in a single build.

### DIFF
--- a/src/main/java/com/palantir/gradle/gitversion/GitVersionArgs.java
+++ b/src/main/java/com/palantir/gradle/gitversion/GitVersionArgs.java
@@ -3,6 +3,7 @@ package com.palantir.gradle.gitversion;
 import com.google.common.base.Preconditions;
 
 import java.util.Map;
+import java.util.Objects;
 
 class GitVersionArgs {
     private static final String PREFIX_REGEX = "[/@]?([A-Za-z]+[/@-])+";
@@ -20,6 +21,23 @@ class GitVersionArgs {
                 "Specified prefix `" + prefix + "` does not match the allowed format regex `" + PREFIX_REGEX + "`.");
 
         this.prefix = prefix;
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) {
+            return true;
+        }
+        if (other == null || getClass() != other.getClass()) {
+            return false;
+        }
+        GitVersionArgs that = (GitVersionArgs) other;
+        return Objects.equals(prefix, that.prefix);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(prefix);
     }
 
     // groovy closure invocation allows any number of args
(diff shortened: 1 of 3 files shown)


src/main/java/com/palantir/gradle/gitversion/GitVersionArgs.java
```diff
@@ -3,6 +3,7 @@
 import com.google.common.base.Preconditions;
 
 import java.util.Map;
+import java.util.Objects;
 
 class GitVersionArgs {
```
raiju, line 8: Should we be pulling in Immutables to auto-gen most of this class? (We currently don't use annotation processors in this project)

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
```diff
@@ -1,31 +1,32 @@
...
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 
-import java.io.File;
-import java.io.IOException;
-
 public class GitVersionPlugin implements Plugin<Project> {
-    public void apply(final Project project) {
+    private final Map<GitVersionArgs, Map<File, VersionDetails>> cachedVersionDetails = new HashMap<>();
```
raiju, line 17: How does this interact with gradle daemon? Is there a way to reset this on every build start, or will the daemon do this automatically? In short, what is the lifecycle of a gradle plugin?

gatesn:
Given https://github.com/palantir/gradle-git-version/pull/96 - maybe we should be careful here

src/main/java/com/palantir/gradle/gitversion/GitVersionArgs.java
dansanduleac: yes!

raiju:
Does anyone know of a way for plugins to be aware of gradle's build lifecycle? It seems not crazy to claim that by default (happy to add a flag that defaults either way) the git version result doesn't change during a single build.

dansanduleac:
> Does anyone know of a way for plugins to be aware of gradle's build lifecycle? It seems not crazy to claim that by default (happy to add a flag that defaults either way) the git version result doesn't change during a single build.

This assumption would be broken if running a publish in continuous mode (`./'gradlew --continuous publishToMavenLocal`), so would have to be a bit careful, e.g. only cache if that mode is off.

PR #150, opened 2019-08-06
Revert changes after 0.11.0

dansanduleac:
## Before this PR
Experimental changes were on develop and we were on RCs for a year.

## After this PR
==COMMIT_MSG==
Revert code back to what was in 0.11.0.
==COMMIT_MSG==


src/main/java/com/palantir/gradle/gitversion/VersionDetails.java
```diff
@@ -59,7 +59,23 @@ private String expensiveComputeRawDescription() {
...
+        // If native failed, return JGit one
+        if (nativeGitDescribe == null) {
+            return jgitDescribe;
+        }
+
+        // If native succeeded, make sure it's same as JGit one
+        if (!nativeGitDescribe.equals(jgitDescribe)) {
+            throw new IllegalStateException(String.format("Inconsistent git describe: native was %s and jgit was %s. " +
+                    "Please report this on github.com/palantir/gradle-git-version", nativeGitDescribe, jgitDescribe));
+        }
```
iamdanfox, line 76: If we're saying we don't need the jgit stuff, then why are we still doing the dual implementation?

src/main/groovy/com/palantir/gradle/gitversion/GitCli.groovy
```diff
@@ -0,0 +1,77 @@
...
+
+        // stop at the root directory, return non-existing File object
+        if (currentRoot.parentFile == null) {
+            return gitDir
+        }
+
+        // look in parent directory
+        return scanForRootGitDir(currentRoot.parentFile)
+    }
+}
```
iamdanfox: It's a bit of a shame this loses all the groovy -> java conversion work we did back here: https://github.com/palantir/gradle-git-version/pull/90/files

dansanduleac: yes. so let's just revert to 0.11 and not invest more time fixing that up / removing the jgit stuff, until we encounter problems with it

src/main/java/com/palantir/gradle/gitversion/VersionDetails.java
dansanduleac: I am not saying that. I'm in favour of reverting this to the way 0.11.0 behaved because that's the last non-rc and it's being used in a lot of places.
We should leave it as it is even if it's using a hybrid of jgit and git, and only attempt to fix it if we encounter problems.

iamdanfox (approved):

PR #152, opened 2019-08-08
Fix publishing to bintray & gradle

dansanduleac:
## Before this PR
Publishing has been broken since #107 since resolved versions were no longer being included in the produced POM.

## After this PR
==COMMIT_MSG==
Fix publishing to bintray and gradle.
==COMMIT_MSG==


dansanduleac:
I have no clue what is breaking this test, it can't find a method that is clearly there. It seems to be looking at the right class too.

versions.props
```diff
@@ -1,5 +1,5 @@
 org.eclipse.jgit:org.eclipse.jgit = 4.5.4.201711221230-r
-com.google.guava:guava = 28.0-jre
+com.google.guava:guava = 23.5-jre
```
dansanduleac, line 2: thought 23.5 should be enough so no need to bump higher

dansanduleac:
Verified this produces a correct POM by running `generatePomFileForPluginMavenPublication` locally

PR #158, opened 2019-09-03
Apply baseline

ferozco:
Apply baseline


build.gradle
```diff
@@ -1,42 +1,29 @@
...
+        maven { url "http://palantir.bintray.com/releases" }
     }
 }
 
 plugins {
     id 'com.jfrog.bintray' version '1.8.4'
     id 'com.gradle.plugin-publish' version '0.9.9'
-    id 'com.palantir.configuration-resolver' version '0.2.0'
     id 'com.palantir.git-version' version '0.4.0'
     id 'com.palantir.idea-test-fix' version '0.1.0'
```
iamdanfox: this is a mystical plugin that we can't find the source for anymore - deletable?

src/main/java/com/palantir/gradle/gitversion/GitVersionArgs.java
```diff
@@ -13,6 +28,7 @@ public String getPrefix() {
         return prefix;
     }
 
+    @SuppressWarnings("PreconditionsConstantMessage")
```
iamdanfox, line 31: https://github.com/palantir/gradle-baseline/issues/389

PR #160, opened 2019-09-04
Fix publish

ferozco:
## Before this PR
We had failed to correctly publish `0.12.0` to gradle, preventing users using the plugin syntax to upgrade to latest

## After this PR
==COMMIT_MSG==
Correctly publish gradle-git-version as a gradle plugin
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,11 +2,14 @@ buildscript {
     repositories {
         maven { url "http://palantir.bintray.com/releases" }
     }
+    dependencies {
+        classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.4'
+        classpath 'com.netflix.nebula:nebula-publishing-plugin:13.2.0'
+    }
 }
 
 plugins {
     id 'com.gradle.plugin-publish' version '0.9.9'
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -99,7 +63,8 @@ pluginBundle {
     }
 }
 
-publishPlugins.onlyIf {
-    System.env.CIRCLE_TAG
-}
+// Configure the publishPlugins task
+tasks.publish.dependsOn publishPlugins
```
dansanduleac, line 67: probably want this here too

```
publishPlugins.onlyIf {
    versionDetails().isCleanTag
}
```

build.gradle
```diff
@@ -2,11 +2,14 @@ buildscript {
     repositories {
         maven { url "http://palantir.bintray.com/releases" }
```
dansanduleac, line 3: http:(

dansanduleac:
It needs a few more fixes sadly, like removing `<<` obsolete gradle syntax which no longer works in v5.
I've fixed up #152 which had these on already, so let's go with that one :)

PR #261, opened 2020-01-10
Added rudimentary support for worktrees.

aioobe:
## Before this PR
The plugin crashes when using worktrees.

## After this PR
==COMMIT_MSG==
Rudimentary support for worktrees added.
==COMMIT_MSG==

## Possible downsides?
JGit doesn't yet support worktrees, so there are still some corner cases unsolved, such as creating a separate worktree out of an empty repository.


aioobe:
@ferozco, do you think you'll have time to take a look at this?

src/main/java/com/palantir/gradle/gitversion/VersionDetails.java
```diff
@@ -97,6 +110,10 @@ private String expensiveComputeRawDescription() {
     }
 
     private boolean isRepoEmpty() {
+        if (isWorktree()) {
+            return false; // not implemented for worktrees yet
```
ferozco, line 114: same here, what causes work trees to be different

src/main/java/com/palantir/gradle/gitversion/NativeGitDescribe.java
```diff
@@ -51,7 +51,7 @@ private String runGitCmd(String... commands) throws IOException, InterruptedExce
         cmdInput.add("git");
         cmdInput.addAll(Arrays.asList(commands));
         ProcessBuilder pb = new ProcessBuilder(cmdInput);
-        pb.directory(directory);
+        pb.directory(directory.getParentFile());
```
ferozco, line 54: why is this necessary?

ferozco (commented):
Apologies for the delay in review. I'm cool with adding support for worktrees but I'm worried that it will be hard to maintain and make the rest of the code harder to understand since we now have an extra variable to account for

src/main/java/com/palantir/gradle/gitversion/NativeGitDescribe.java
aioobe: Here `directory` refers to the `.git` directory. In a worktree `.git` is a file (containing the path to the actual `.git` directory). Adding `.getParentFile()` gets us up to the repository root, which is always an actual directory and just as good of a directory to execute git commands from.

src/main/java/com/palantir/gradle/gitversion/VersionDetails.java
aioobe: The jgit implementation of this method is poorly implemented, and will throw an exception (and then return `true`) in case we're in a worktree directory. We could perhaps try to come up with a native command for `isRepoEmpty`, but the risk is that native `git` isn't available in the environment, so I took the safe route here for now.

aioobe:
> Apologies for the delay in review. I'm cool with adding support for worktrees but I'm worried that it will be hard to maintain and make the rest of the code harder to understand since we now have an extra variable to account for

I agree that it's a bit messy. FWIW I asked what the status was on worktree support in jgit [over here](https://bugs.eclipse.org/bugs/show_bug.cgi?id=477475#c42) and it seems things are moving in the right direction. Once such support is in place, I expect us to be able to revert this fix and replace it with a clean jgit implementation.

aioobe:
beep.

src/main/java/com/palantir/gradle/gitversion/NativeGitDescribe.java
Robinhola: I like this. Maybe now is a good time to rename or add documentation around this `directory` variable.

PR #329, opened 2020-03-26
Performance stats in build scans

CRogers:
## Before this PR
We suspect this plugin is the cause of many Gradle projects being very slow to evaluate, so having long pauses before doing anything useful. Before I go and speed it up, it would be useful to expose this information in build scans so we can see if my future work will actually speed it up.

## After this PR
==COMMIT_MSG==
Performance information is exposed in Gradle build scans for `gitVersion()` and `versionDetails().xxxxx()` calls.
==COMMIT_MSG==

You can now see what's happening in the build scan (even per project):

Screen Shot 2020-03-26 at 3 47 57 PM

## Possible downsides?
More complexity. It's impossible to test that we've actually set the values in the build scan (so I just test the build doesn't crash when for various configurations)


build.gradle
```diff
@@ -52,6 +60,14 @@ tasks.withType(JavaCompile) {
     options.compilerArgs += ['-Werror', '-Xlint:deprecation']
 }
 
+// Access groovy from java sourceset
+sourceSets {
+    main {
+        java { srcDirs = [] }
+        groovy { srcDirs += ['src/main/java'] }
+    }
+}
```
dansanduleac: ooh hacky :P

src/main/groovy/com/palantir/gradle/gitversion/BuildScanPluginInterop.groovy
```diff
@@ -0,0 +1,44 @@
...
+
+package com.palantir.gradle.gitversion
+
+import java.util.function.Supplier
+import org.gradle.api.Project
+
+class BuildScanPluginInterop {
+    static void addBuildScanCustomValues(Project rootProject, Supplier<Map<String, String>> customValues) {
+        // After evaluate because while we can detect the <5.x com.gradle.build-scan project on the root project,
+        // there is no way to detect the >6.x com.gradle.enterprise settings plugins using withPlugin
```
dansanduleac: can't we do `project.gradle.pluginManager.withPlugin`?

CRogers: It's on `Settings`, not `Gradle`. There is no public way to get `Settings` -you have to do:

```java
Settings settings = ((org.gradle.api.internal.GradleInternal)
                        getProject().getGradle())
                .getSettings();
```

build.gradle
CRogers: I thought this might catch your eye, but I didn't really want to move all the existing code into `groovy`, at least not in this PR.

dansanduleac: It just occurred to me that doing things this way means we don't run errorprone on anything (and we should probably fix it in other places where we do this).
Can we put this groovy file in a different source set perhaps that we expose through main?


```groovy
sourceSets {
  groovy
  main.output.dir groovy.output
}

// one/both of these might not be necessary
configurations.runtimeOnly.extendsFrom configurations.groovyRuntimeClasspath
configuration.implementation.extendsFrom configurations.groovyCompile
```

src/main/groovy/com/palantir/gradle/gitversion/BuildScanPluginInterop.groovy
dansanduleac: and perhaps not write this in groovy?

edit: yep I understand why this is easier than using reflection, maybe it's ok for now

CRogers: I do `gradleEnterprise.buildScan` and other lovely stuff like that in this though.

CRogers:
:+1: approving dan's changes

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -596,13 +596,89 @@ class GitVersionPluginTests extends Specification {
...
-            .withPluginClasspath()
-            .withProjectDir(projectDir)
-            .withArguments(arguments)
+
+        def gradleRunner = GradleRunner.create()
+                .withPluginClasspath()
+                .withProjectDir(projectDir)
+                .withArguments(arguments)
+
+        gradleVersion.ifPresent({ version -> gradleRunner.withGradleVersion(version) })
```
dansanduleac, line 679: we're duplicating a bunch of logic from IntegrationTestKitSpec here, maybe we should just change to that

CRogers: I want to change them all to nebula, so I can actually run them from my IDE

CRogers: But that's a task for a later PR

dansanduleac (approved):

ilka-schulz:
The test `does not crash when setting build scan custom values when Gradle 5 build scan plugin is applied` which has been introduced in [this commit](https://github.com/palantir/gradle-git-version/blob/22842eb18a2e6c26de1837bd4b2810c7d646af98/src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy#L598) runs for roughly four minutes and then does not pass on my machine:

```text
GitVersionPluginTests > does not crash when setting build scan custom values when Gradle 5 build scan plugin is applied FAILED
    org.gradle.testkit.runner.UnexpectedBuildFailure at GitVersionPluginTests.groovy:614

32 tests completed, 1 failed
```

my setup:

```text
$ gradle --version

------------------------------------------------------------
Gradle 6.9
------------------------------------------------------------

Build time:   2021-05-07 07:28:53 UTC
Revision:     afe2e24ababc7b0213ccffff44970aa18035fc0e

Kotlin:       1.4.20
Groovy:       2.5.12
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          14 (Oracle Corporation 14+36-1461)
OS:           Linux 5.4.125-1.fc25.qubes.x86_64 amd64
```

Is it me or is this test flawed?

**Update:** Never mind, it works with OpenJDK 11

PR #342, opened 2020-04-21
fix: allow tagging the same commit with multiple tags

close2:
Fixes "Inconsistent git describe" errors in case multiple tags with different prefixes reference the same commit.

When using prefixes the same commit might have different tags.

The jgit code did not take the prefix into account and stored tags in a map:
commit → tag

If multiple tags with different prefixes referenced the same commit only the last
tag was stored in this map.

This commit takes a prefix into account and ignores all tags which do not start
with the given prefix.

### DIFF
--- a/src/main/java/com/palantir/gradle/gitversion/JGitDescribe.java
+++ b/src/main/java/com/palantir/gradle/gitversion/JGitDescribe.java
@@ -16,11 +16,6 @@
 
 package com.palantir.gradle.gitversion;
 
-import java.io.IOException;
-import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-git-version](https://github.com/palantir/gradle-git-version), @close2! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

close2:
@iamdanfox Please take a look at the PR.

PR #462, opened 2021-02-09
Allow numbers in prefix argument

twistedtwin:
Allow numbers in prefix

## Before this PR
Tags with numbers in them are rejected as prefixes.  For example, p123@ would be rejected.

## After this PR
Tags with number them are not rejected.  Prefixes that start with a number would still be rejected.

## Possible downsides?
I'm unable to build and test this, so I'm not sure if the numbers cause an issue downstream.


palantirtech:
Thanks for your interest in [palantir/gradle-git-version](https://github.com/palantir/gradle-git-version), @twistedtwin! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

Issue #353, opened 2020-05-13
afterEvaluate deprecated in Gradle 6; will not work in 7+


gillius:
https://docs.gradle.org/6.3/userguide/upgrading_version_5.html#calling_project_afterevaluate_on_an_evaluated_project_has_been_deprecated

Stacktrace when using Gradle 6.3 and gradle-git-version 0.12.3

```
Using method Project.afterEvaluate(Closure) when the project is already evaluated has been deprecated. This will fail with an error in Gradle 7.0. The configuration given is ignored because the project has already been evaluated. To apply this configuration, remove afterEvaluate. Consult the upgrading guide for further information: https://docs.gradle.org/6.3/userguide/upgrading_version_5.html#calling_project_afterevaluate_on_an_evaluated_project_has_been_deprecated
        at org.gradle.api.internal.project.DefaultProject.maybeNagDeprecationOfAfterEvaluateAfterProjectIsEvaluated(DefaultProject.java:1031)
        at org.gradle.api.internal.project.DefaultProject.afterEvaluate(DefaultProject.java:1021)
        at org.gradle.api.Project$afterEvaluate.call(Unknown Source)
        at com.palantir.gradle.gitversion.BuildScanPluginInterop.addBuildScanCustomValues(BuildScanPluginInterop.groovy:26)
        at com.palantir.gradle.gitversion.GitVersionRootPlugin.apply(GitVersionRootPlugin.java:34)
        at com.palantir.gradle.gitversion.GitVersionRootPlugin.apply(GitVersionRootPlugin.java:26)
```

yatsvic:
From Gradle v7.0.0-M3
```* What went wrong:
An exception occurred applying plugin request [id: 'com.palantir.git-version']
> Failed to apply plugin class 'com.palantir.gradle.gitversion.GitVersionRootPlugin'.
   > Cannot run Project.afterEvaluate(Closure) when the project is already evaluated.

**Comment by tunix** [comment]
gradle 7.0.2 is now the stable version.

**Comment by JohnAmican** [comment]
Getting the same issue with gradle 7.0.2:

```
* What went wrong:
An exception occurred applying plugin request [id: 'com.palantir.git-version', version: '0.12.3']
> Failed to apply plugin class 'com.palantir.gradle.gitversion.GitVersionRootPlugin'.
   > Cannot run Project.afterEvaluate(Closure) when the project is already evaluated.
```

Let me know if you need more detail.

xianzheTM:
[...]

PR #541, opened 2021-12-02
[Fix] #353: Evaluate Buildscan Directly When Root is Evaluated

bmarcaur:
## Before this PR
Issue is described at length: https://github.com/palantir/gradle-git-version/issues/353

## After this PR
Usages of `git-version` on submodules no longer errors out.

## Possible downsides?
It's possible that this [check](https://github.com/palantir/gradle-git-version/blob/develop/src/main/java/com/palantir/gradle/gitversion/GitVersionRootPlugin.java#L29) was meant to fail when a user does not apply the plugin at the root, but as it stands all it does is runtime check that our code properly calls out to a project's root and then applies it. So an alternative fix is to hard error when this is called in submodules. However, fixing that "bug" is a breaking change.


bmarcaur:
Note for the reviewers: I wrote the test, observed the failing behavior, and then I applied the fix. So this definitely exhibits and fixes the behavior described in #353

changelog/@unreleased/pr-541.v2.yml
```diff
@@ -0,0 +1,15 @@
...
+    [Fix] 353: Evaluate Buildscan Directly When Root is Evaluated
+
+    ## Before this PR
+    Issue is described at length: https://github.com/palantir/gradle-git-version/issues/353
+
+    ## After this PR
+    Usages of `git-version` on submodules no longer errors out.
+
+    ## Possible downsides?
+    It's possible that this [check](https://github.com/palantir/gradle-git-version/blob/develop/src/main/java/com/palantir/gradle/gitversion/GitVersionRootPlugin.java#L29) was meant to fail when a user does not apply the plugin at the root, but as it stands all it does is runtime check that our code properly calls out to a project's root and then applies it. So an alternative fix is to hard error when this is called in submodules. However, fixing that "bug" is a breaking change.
```
carterkozak:
```suggestion
    Evaluate Buildscan directly when the root project is evaluated
```

src/groovy/groovy/com/palantir/gradle/gitversion/BuildScanPluginInterop.groovy
```diff
@@ -21,24 +21,36 @@ import org.gradle.api.Project
...
 class BuildScanPluginInterop {
     static void addBuildScanCustomValues(Project rootProject, Supplier<Map<String, String>> customValues) {
-        // After evaluate because while we can detect the <5.x com.gradle.build-scan project on the root project,
-        // there is no way to detect the >6.x com.gradle.enterprise settings plugins using withPlugin
-        rootProject.afterEvaluate {
-            def gradleEnterpriseExtension = rootProject.extensions.findByName('gradleEnterprise')
+        // Fix #353: Detect when the root project is already evaluated, thus anything is afterEvaluate
+        // and we can just execute immediately. This is due to afterEvaluate changes in Gradle 7.
+        //
+        // Note: We cannot use hasCompleted() because we need to support gradle 5
```
carterkozak: > we need to support gradle 5

Why?

bmarcaur: For no other reason that I did not want to take away behavior as I suspected that would require a major version bump. The previous comments seemed to imply that having this work on Gradle 5 was something to be desired? IDC TBH

bmarcaur: I clarified the language, not sure that the upside is worth the loss of gradle 5 at this point.

carterkozak (approved):

Issue #468, opened 2021-03-18
Plugin causes the build to hang with no other information logged


rherrick:
## What happened?

On a new Jenkins server, a couple of our builds that work locally and on an old Jenkins server were hanging. I looked into it and narrowed the cause down to the the **com.palantir.git-version** plugin: if I remove the references to this plugin, the build works fine. If I include this plugin (even if I don't reference any tasks or methods from the plugin), it hangs and just cycles on the daemon waiting to acquire a lock, acquiring a lock, then releasing the lock (including when I run the build with the **--no-daemon** flag specified explicitly and/or **org.gradle.daemon** set to **false** in **~/.gradle/gradle.properties**.

> After I had started writing this issue up, we discovered the critical factor that causes this failure. I don't know _why_ this critical factor causes this failure, so I'm still reporting it. This plugin fails as described below whenever the user running the build is authenticated via an external repo (LDAP) and/or when the project being built is located somewhere that's mounted to the server, in our case from ZFS storage via NFS. Tomorrow I'll try to determine whether the fault is with the externally authenticated user or with the network-mounted storage location (I need some IT help to get a scenario in which I can test this set up) and add to this report.

For reference, the builds that we're having trouble with are the **master** branches of these two repos:

* https://bitbucket.org/xnatdev/xnat-data-models
* https://bitbucket.org/xnatdev/xnat-web

I've found a few clues that may help and tried a few things that _didn't_ help:

* We tried different versions of git on the system. The Jenkins server is running on CentOS 7.9, which includes git 1.8.3.1. I thought maybe upgrading git would work, but cloning the repo with git 2.17.1 didn't change anything.
* We have the same version of Java on different servers where the build works.
* I spun up a Vagrant VM with CentOS 7.9 (same version of git and Java) and built there with no problems.
* The issue isn't limited to running within Jenkins: I cloned https://bitbucket.org/xnatdev/xnat-data-models and tried to build the **master** branch in my own home folder. Same result as running in Jenkins (hang).
* I tried with daemon on and parallel true and with no daemon or parallel. No change.

I ran these builds both on the Jenkins server (not under Jenkins, just on the same machine) and locally with debug logging on and captured the output. You can find both the raw log files here, as well as the logs filtered down to just show lines that include references to _git_, _palantir_, or _version_, in order to better focus on what's actually happening with the plugin:

https://gist.github.com/rherrick/9894271ad99ba50fc405ae2a4d3d870f

[...]

PR #581, opened 2022-03-21
Bumped dependency versions

mikethecalamity:
## Before this PR
Issue is described at length: https://github.com/palantir/gradle-git-version/issues/468

## After this PR
Usages of `git-version` on submodules no longer errors out.

## Possible downsides?
New versions have breaking changes to the used APIs.


palantirtech:
Thanks for your interest in [palantir/gradle-git-version](https://github.com/palantir/gradle-git-version), @mweber03! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

versions.props
```diff
@@ -1,11 +1,11 @@
-org.eclipse.jgit:org.eclipse.jgit = 5.3.2.201906051522-r
+org.eclipse.jgit:org.eclipse.jgit = 6.1.0.202203080745-r
 com.google.guava:guava = 31.1-jre
-org.immutables:* = 2.8.8
+org.immutables:* = 2.9.0
```
carterkozak: This version has a known memory leak: https://github.com/immutables/immutables/issues/1355

Let's bump only jgit in this change, and see what that does for us. A `./gradlew --write-locks` will be necessary to update the `versions.lock`

mikethecalamity: done and done

carterkozak (approved):

PR #684, opened 2023-01-30
add caching to git describe

yiweny:
## Before this PR
`gitVersion()` or `versionDetails()` does not have caching, every time it is called, it will create a new 'VersionDetails' object and call the expensive `git describe` operation to get the version. This causes long configuration time in a lot of the repos where `version gitVersion()` is called under allprojects. 

## After this PR
Build service is used so if we call `gitVersion()` on all projects, we will not call the expensive `git describe` many times. Instead, the build service will have a concurrent hash map that stores the created `VersionDetails` object with a specific prefix as key the first time it is called. If `gitVersion()` is previously computed, it will return the cached value the next time it is called.

## Possible downsides?
Build service is only supported from Gradle 6. This means this plugin will no longer support Gradle 5. 
</changelog-description><changelog-links></changelog-links><changelog-tags></changelog-tags>


src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -611,6 +611,98 @@ class GitVersionPluginTests extends Specification {
...
+        git.tag().setAnnotated(true).setName('1.0.0').setMessage('1.0.0').call()
+        BuildResult buildResult1 = with(':submodule1:printVersion').build()
+
+        when:
+        BuildResult buildResult2 = with(':submodule2:printVersion').build()
+
+        then:
+        buildResult1.output.contains(":printVersion\n1.0.0\n")
+        buildResult2.output.contains(":printVersion\n1.0.0\n")
+    }
```
yiweny: I'd like to explain about this test case a little bit. The build file of `submodule2` makes sure that before we are calling `:submodule2:printVersion`, we change the latest tag to be `2.0.0` in the work tree. However, since the build service will always be registered at configuration time by root project(at a time that this git command isn't executed) and the call to `printVersion` on `submodule2` is always cached, the result of `:submodule2:printVersion` would be `1.0.0`. 

The thing that I am worried about is, even though this test case showcases that caching is indeed successful, does it also imply a potential bug/issue with the logic? Suppose we have a build file like `subModuleBuildFile2` where the git status of the repo changes during execution time. Then the correctness of the plugin cannot be guaranteed unless we add a bunch of expensive operations to check if the tags have been modified.

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
```diff
@@ -16,38 +16,56 @@
...
 
     @Override
     public void apply(final Project project) {
-        project.getRootProject().getPluginManager().apply(GitVersionRootPlugin.class);
+        int maxParallelUsage = 2;
+        Provider<GitVersionCacheService> serviceProvider = project.getGradle()
+                .getSharedServices()
+                .registerIfAbsent("GitVersionCacheService", GitVersionCacheService.class, spec -> {
+                    // Provide some parameters
+                    spec.getMaxParallelUsages().set(maxParallelUsage);
```
CRogers: Do we need to set a max parallel usages at all? I think provided we make the class thread safe, it should be ok to allow max parallel access.

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
```diff
@@ -16,38 +16,56 @@
...
+                });
 
-        final Git git = gitRepo(project);
+        if (project.getRootProject() == project) {
```
CRogers: I think it'd be best to leave this in the root plugin, if possible. Just to separate out the code a bit and avoid this if.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -0,0 +1,101 @@
...
+        File gitDir = getRootGitDir(new File(project));
+        GitVersionArgs gitVersionArgs = GitVersionArgs.fromGroovyClosure(args);
+        String key = gitDir.toString() + "|" + gitVersionArgs.getPrefix();
+        if (versionDetailsMap.containsKey(key)) {
+            return versionDetailsMap.get(key).getVersion();
+        }
+        Git git = gitRepo(gitDir);
+        VersionDetails versionDetails =
+                TimingVersionDetails.wrap(timer, new VersionDetailsImpl(git, GitVersionArgs.fromGroovyClosure(args)));
+        versionDetailsMap.put(key, versionDetails);
```
CRogers: This is not thread safe. I'd make the map a `ConcurrentHashMap` and use `.computeIfAbsent`:

```java
versionDetailsMap.computeIfAbsent(key, _ignored -> {
    // calculate version details for key
});
```

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public GitVersionCacheService() {
+        timer = new Timer();
+        versionDetailsMap = new HashMap<>();
+    }
+
+    public final String getGitVersion(String project, Object args) {
+        File gitDir = getRootGitDir(new File(project));
+        GitVersionArgs gitVersionArgs = GitVersionArgs.fromGroovyClosure(args);
+        String key = gitDir.toString() + "|" + gitVersionArgs.getPrefix();
```
CRogers: Why does git dir need to be part of the key?

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -0,0 +1,101 @@
...
+        if (versionDetailsMap.containsKey(key)) {
+            return versionDetailsMap.get(key).getVersion();
+        }
+        Git git = gitRepo(gitDir);
+        VersionDetails versionDetails =
+                TimingVersionDetails.wrap(timer, new VersionDetailsImpl(git, GitVersionArgs.fromGroovyClosure(args)));
```
CRogers: ~~I think reusing the `Timer` field here is not thread safe either - I'd make a new `Timer` object each time here as a local variable.~~

Ignore this, I forgot this was a custom type we made that is thread safe.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -0,0 +1,101 @@
...
+        versionDetailsMap.put(key, versionDetails);
+        String gitVersion = versionDetails.getVersion();
+        return gitVersion;
+    }
+
+    public final VersionDetails getVersionDetails(String project, Object args) {
+        File gitDir = getRootGitDir(new File(project));
```
CRogers: I would pass the project dir as a File rather than a String here.

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
CRogers: I think assuming that the underlying state of the git tree does not change during a Gradle execution is reasonable. The thing we want to make sure is we do not cache the version details between Gradle executions.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -0,0 +1,101 @@
...
+import org.gradle.api.services.BuildServiceParameters;
+
+public abstract class GitVersionCacheService implements BuildService<BuildServiceParameters.None> {
+
+    private final Timer timer;
+    private Map<String, VersionDetails> versionDetailsMap;
+
+    public GitVersionCacheService() {
+        timer = new Timer();
+        versionDetailsMap = new HashMap<>();
```
CRogers: You can just join the declaration and assignment.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Map;
+import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.internal.storage.file.FileRepository;
+import org.gradle.api.services.BuildService;
+import org.gradle.api.services.BuildServiceParameters;
+
+public abstract class GitVersionCacheService implements BuildService<BuildServiceParameters.None> {
+
+    private final Timer timer;
+    private Map<String, VersionDetails> versionDetailsMap;
```
CRogers: This can be final

changelog/@unreleased/pr-684.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: add caching to git describe
```
CRogers: We need a better description here. I'd make it a `break` and say it removes support for Gradle 5 (and say what is the new minimum version we support). I'd also a bit more detail, saying how the pattern of calling `gitversion()` in each project used to lead to extremely long configuration times, however these are now cached so it's faster.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
bjlaub: We already have guava as a dependency here - perhaps just a `LoadingCache`?

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public final String getGitVersion(String project, Object args) {
+        File gitDir = getRootGitDir(new File(project));
+        GitVersionArgs gitVersionArgs = GitVersionArgs.fromGroovyClosure(args);
+        String key = gitDir.toString() + "|" + gitVersionArgs.getPrefix();
+        if (versionDetailsMap.containsKey(key)) {
+            return versionDetailsMap.get(key).getVersion();
+        }
+        Git git = gitRepo(gitDir);
```
bjlaub: I wonder if this could just be pushed down into `VersionDetailsImpl`? For a couple of reasons:
1. It seems odd to initialize a JGit object in this class just to inject it into the VersionDetailsImpl instance
2. If we ever do decide to eliminate JGit, it would be good to have the usage limited in scope to make refactoring easier

On the _other hand_, I wonder if the `Git` class is thread-safe? The JGit javadoc doesn't really specify and a quick search indicates... maybe? If it is, then I wonder if it would make sense to cache a single instance here in this class instead of creating one instance per VersionDetailsImpl.

My preference would probably be to move this inside `VersionDetailsImpl` and instead pass `gitDir` as a constructor arg; a quick glance at `VersionDetailsImpl` seems to indicate that it is using _both_ JGit and Native Git to compute the description, which seems a little broken (see e.g.: https://github.com/palantir/gradle-git-version/blob/develop/src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java#L54-L60)

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
CRogers: A cache that expires is unnecessary complexity for this use case imo, the number of items in this map should never grow too large, and evicting entries is probably not what we want.

bjlaub: Ah well, the checkstyle config bans guava caches in favor of Caffeine anyway :D

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
yiweny: Because I am thinking of the scenario where different projects have different work tree. I can remove this if it is considered an extreme edge case not likely to happen.

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
yiweny: Cool. I can delete the new test cases I wrote.

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
yiweny: I removed this line.

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
yiweny: I added back the `GitVersionRootPlugin`. The only reason why I deleted the root plugin is that we'd have to have the following in both plugins 
```
        Provider<GitVersionCacheService> serviceProvider = project.getGradle()
                .getSharedServices()
                .registerIfAbsent("GitVersionCacheService", GitVersionCacheService.class, _spec -> {});
```

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
yiweny: I resolved by using the concurrent hash map instead of hash map here.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
yiweny: ^^ I removed the `gitDir` part in the key. If you think the above case might happen and it is necessary to keep the worktree as part of the key, please let me know.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
yiweny: Fixed.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
yiweny: Fixed

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
yiweny: Fixed

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
yiweny: Fixed by passing `gitDir` to the constructor rather than `Git` object. @bjlaub Do you want me to delete the Native git part as well?

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
```diff
@@ -17,37 +17,32 @@
...
-    private final Timer timer = new Timer();
 
     @Override
     public void apply(final Project project) {
         project.getRootProject().getPluginManager().apply(GitVersionRootPlugin.class);
 
-        final Git git = gitRepo(project);
+        Provider<GitVersionCacheService> serviceProvider = project.getGradle()
+                .getSharedServices()
+                .registerIfAbsent("GitVersionCacheService", GitVersionCacheService.class, _spec -> {});
```
CRogers: I would make a static method to get the `Provider<GitVersionCacheService>` from a `Project` which is shared by both plugins.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -0,0 +1,87 @@
...
+import org.gradle.api.services.BuildServiceParameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class GitVersionCacheService implements BuildService<BuildServiceParameters.None> {
+
+    private static final Logger log = LoggerFactory.getLogger(GitVersionCacheService.class);
+
+    private final Timer timer = new Timer();
+    private final Map<GitVersionArgs, VersionDetails> versionDetailsMap = new ConcurrentHashMap<>();
```
CRogers: I would make this a `ConcurrentMap` to make it obvious it's the concurrent version.

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
CRogers: You could even put it on `GitVersionCacheService`

changelog/@unreleased/pr-684.v2.yml
```diff
@@ -0,0 +1,16 @@
+type: break
+break:
+  description: "## Before this PR\n`gitVersion()` or `versionDetails()` does not have
+    caching, every time it is called, it will create a new 'VersionDetails' object
+    and call the expensive `git describe` operation to get the version. This causes
+    long configuration time in a lot of the repos where `version gitVersion()` is
+    called under allprojects. \n\n## After this PR\nBuild service is used so if we
+    call `gitVersion()` on all projects, we will not call the expensive `git describe`
+    many times. Instead, the build service will have a concurrent hash map that stores
+    the created `VersionDetails` object with a specific prefix as key the first time
```
CRogers: I think this sentence is too much detail for an end user - I'd leave out any details of how the caching happens (build service, hashmaps, prefixes etc) and just say it's now cached.

changelog/@unreleased/pr-684.v2.yml
```diff
@@ -0,0 +1,16 @@
+type: break
+break:
+  description: "## Before this PR\n`gitVersion()` or `versionDetails()` does not have
```
CRogers: I would also leave out the "Before this PR" bits and just have a couple of bullet points, one for the change and one describing the break.

src/main/java/com/palantir/gradle/gitversion/NativeGitDescribe.java
```diff
@@ -1,126 +0,0 @@
-/*
```
bjlaub: out of curiosity, what's the reason for removing this? If there's a potential future where we remove jgit, don't we want to keep this code around?

yiweny: Yeah makes sense. I deleted because it won't be used if we do not call Native git.

bjlaub: > because it won't be used if we do not call Native git.

not sure i follow - don't we _always_ try native git first, and then fall back to jgit if it fails?

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -0,0 +1,95 @@
...
+import org.gradle.api.services.BuildServiceParameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class GitVersionCacheService implements BuildService<BuildServiceParameters.None> {
+
+    private static final Logger log = LoggerFactory.getLogger(GitVersionCacheService.class);
+
+    private final Timer timer = new Timer();
+    private final ConcurrentMap<GitVersionArgs, VersionDetails> versionDetailsMap = new ConcurrentHashMap<>();
```
CRogers: I tried this out on a project and was wondering why it doesn't appear to produce a speedup on repos with large numbers of projects and calling `gitversion()` on each. I think this is the reason - `GitVersionArgs` does not implement `hashCode()` or `equals()` so a new entry is created on the map each time. I'd either just use `String` here (probably easiest), or make a new immutables type containing the string prefix and immutables will generate `hashCode()` and `equals()` for us.

yiweny: I switched back to using the concatenation of gitDir path and prefix as the key.

yiweny: If we do not use the gitDir as part of the key, we'd fail the `test subproject version` test where the two projects have different worktree.

CRogers (approved):
Looks good!

Issue #564, opened 2022-02-11
Git root dir does not work for worktrees


blakehurlburt:
When using a worktree, gradle tasks run into the following error:
```
java.nio.file.FileSystemException: <worktree directory>/.git/config: Not a directory
        at com.palantir.gradle.gitversion.GitVersionPlugin.gitRepo(GitVersionPlugin.java:73)
        at com.palantir.gradle.gitversion.GitVersionPlugin.apply(GitVersionPlugin.java:36)
        at com.palantir.gradle.gitversion.GitVersionPlugin.apply(GitVersionPlugin.java:29)
```

Similar for
```
java.nio.file.FileSystemException: <worktree directory>/.git/refs/HEAD: Not a directory
        at com.palantir.gradle.gitversion.VersionDetailsImpl.isRepoEmpty(VersionDetailsImpl.java:102)
        at com.palantir.gradle.gitversion.VersionDetailsImpl.expensiveComputeRawDescription(VersionDetailsImpl.java:74)
        at com.palantir.gradle.gitversion.VersionDetailsImpl.description(VersionDetailsImpl.java:67)
        at com.palantir.gradle.gitversion.VersionDetailsImpl.getVersion(VersionDetailsImpl.java:47)

        at com.palantir.gradle.gitversion.VersionDetailsImpl.isRepoEmpty(VersionDetailsImpl.java:102)
        at com.palantir.gradle.gitversion.VersionDetailsImpl.expensiveComputeRawDescription(VersionDetailsImpl.java:74)
        at com.palantir.gradle.gitversion.VersionDetailsImpl.description(VersionDetailsImpl.java:67)
        at com.palantir.gradle.gitversion.VersionDetailsImpl.getVersion(VersionDetailsImpl.java:47)
```

https://github.com/palantir/gradle-git-version/blob/26eada382f2327290424e8ebc0c5b1fe818325e6/src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java#L87-L101

This seems to assume the `.git` file in the repo is always a git dir, which is not true for worktrees.

The correct solution is to use the command `git rev-parse --git-dir` for the worktree-specific git dir, and `git rev-parse --git-common-dir` for the shared git dir for all worktrees in the checkout.

drgeb:
I am observing the same issue! Is there a workaround for this ?

theothergp:
This exact issue happens using git submodules: https://git-scm.com/book/en/v2/Git-Tools-Submodules

[...]

Issue #65, opened 2017-07-24
support git worktree structure


jboreiko:
Currently the gradle-git-version plugin will return unspecified if you are using a directory that relies on [git worktrees](https://git-scm.com/docs/git-worktree).  

Unfortunately support for worktrees doesn't yet exist in JGit so I'm afraid we may have to wait a bit longer.  There was some movement on the issue recently - https://bugs.eclipse.org/bugs/show_bug.cgi?id=477475 so maybe will have some luck.

Andy-L:
What is the symptom?  I'm getting:
  "Cannot invoke method getObjectId() on null object" 
when I invoke `versionDetails()` in my `build.gradle`, and am wondering if it is the same issue.

Andy-L:
and the stacktrace looks like:
```
Caused by: java.lang.NullPointerException: Cannot invoke method getObjectId() on null object
        at com.palantir.gradle.gitversion.GitVersionPlugin.memoizedMethodPriv$gitHashProject(GitVersionPlugin.groovy:86)
        at com.palantir.gradle.gitversion.GitVersionPlugin$_closure4.doCall(GitVersionPlugin.groovy)
        at com.palantir.gradle.gitversion.GitVersionPlugin.gitHash(GitVersionPlugin.groovy)
        at com.palantir.gradle.gitversion.GitVersionPlugin.memoizedMethodPriv$versionDetailsProject(GitVersionPlugin.groovy:53)
        at com.palantir.gradle.gitversion.GitVersionPlugin$_closure1.doCall(GitVersionPlugin.groovy)
        at com.palantir.gradle.gitversion.GitVersionPlugin.versionDetails(GitVersionPlugin.groovy)
        at com.palantir.gradle.gitversion.GitVersionPlugin$_apply_closure8.doCall(GitVersionPlugin.groovy:38)
        at com.palantir.gradle.gitversion.GitVersionPlugin$_apply_closure8.doCall(GitVersionPlugin.groovy)
        at org.gradle.internal.metaobject.BeanDynamicObject$MetaClassAdapter.invokeMethod(BeanDynamicObject.java:479)
        at org.gradle.internal.metaobject.BeanDynamicObject.tryInvokeMethod(BeanDynamicObject.java:191)
        at org.gradle.internal.metaobject.MixInClosurePropertiesAsMethodsDynamicObject.tryInvokeMethod(MixInClosurePropertiesAsMethodsDynamicObject.java:42)
[...]

PR #685, opened 2023-01-31
remove jgit

yiweny:
## Before this PR
The plugin uses JGit library to perform git operations.

## After this PR
The plugin removes all usages of JGit and runs git commands with Java ProcessBuilder.

## Possible downsides?
If Git is not installed in the project, the job would fail.


src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -47,63 +46,25 @@ public String getVersion() {
...
-        if (isRepoEmpty()) {
-            log.debug("Repository is empty");
-            return null;
-        }
-
-        String nativeGitDescribe = new NativeGitDescribe(git.getRepository().getDirectory()).describe(args.getPrefix());
-        String jgitDescribe = new JGitDescribe(git).describe(args.getPrefix());
-
-        // If native failed, return JGit one
-        if (nativeGitDescribe == null) {
```
CRogers, line 85: Hmmm, this actually ends up using the jgit version if the native one fails - I misremembered and thought that it was always using the native one. I think we need to think about this carefully - actually removing jgit will cause this to fail in environments where the `git` binary is not installed.

src/main/java/com/palantir/gradle/gitversion/NativeGit.java
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.gitversion;
+
+import java.util.Map;
+
+interface NativeGit {
```
pkoenig10: If we only support native git, do we really need the "native" qualifier?

```suggestion
interface Git {
```

yiweny: @pkoenig10 Thanks for your review. However, currently we cannot merge this because there might be situations on highside where `git` binary is not installed. We cannot risk failure.

src/main/java/com/palantir/gradle/gitversion/GitImpl.java
```diff
@@ -25,33 +26,55 @@
...
     private static final Splitter WORD_SPLITTER = Splitter.on(" ").omitEmptyStrings();
 
     private final File directory;
 
-    NativeGitDescribe(File directory) {
+    GitImpl(File directory) {
+        if (!gitCommandExists()) {
+            throw new RuntimeException("Git not found in project");
+        }
+        this.directory = directory;
```
CRogers: I would replace this whole block with `this(directory, false)` rather than repeating code.

changelog/@unreleased/pr-685.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
```
CRogers: I would make this a `break` so we increment the minor version number and say in the description that the native `git` binary is now required.

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.gitversion;
+
+import java.util.Map;
+
+interface Git {
```
CRogers: Since there's only one implementation of this interface, and we don't actually pass it around, I'd get rid of this interface entirely and just have the implementation.

src/main/java/com/palantir/gradle/gitversion/GitImpl.java
```diff
@@ -25,33 +26,55 @@
...
 import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * Mimics git describe by using rev-list to support versions of git < 1.8.4.
  */
-class NativeGitDescribe implements GitDescribe {
-    private static final Logger log = LoggerFactory.getLogger(NativeGitDescribe.class);
+class GitImpl implements Git {
```
CRogers: I would call this `Git` or `NativeGit` rather than having `Impl` in the name.

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
CRogers: Discussed internal and it makes sense to get rid of jgit

changelog/@unreleased/pr-685.v2.yml
CRogers: Could also mention the benefits like:

* Faster to calculate git version
* Worktrees are now supported??? (maybe we can close #564 and #65?)

pkoenig10: This PR should close those worktree issues.

However we are still blocked on using worktrees because our publishing plugins (both internal and external) still use jgit because they depend on https://github.com/nebula-plugins/gradle-info-plugin.

yiweny: Thanks for the comments. I have updated the changelog and also added a test to show that git worktree is indeed now supported. https://github.com/palantir/gradle-git-version/pull/698

src/main/java/com/palantir/gradle/gitversion/Git.java
yiweny: Thanks. I resolved by removing the interface and just having the implementation.

CRogers (approved):

PR #688, opened 2023-02-06
remove test case related to gradle 5

yiweny:
## Before this PR
Gradle5 does not support build service. So if we want to support build service, we cannot support gradle 5.

## After this PR
Gradle 5 won't be supported

## Possible downsides?
current other plugins, e.g. gradle-baseline-java explicitly doesn't support anything less than gradle 7. So we should be fine not supporting gradle 5 here.


CRogers:
Provided the changelog file is removed, this is good to merge. We can add a changelog about Gradle 5 support in the PR that removes support.

CRogers:
:+1:

PR #697, opened 2023-03-02
Esword/custom version formatting

esword:
## Before this PR
Projects that use git information for versions but have slightly different formatting cannot use the plugin.

## After this PR

==COMMIT_MSG==
Allow customization of the version string.
==COMMIT_MSG==


esword:
just tagging as do not merge because I may add another test to demonstrate how to use in the full plugin

esword:
went with a different solution

Issue #704, opened 2023-03-28
Use git-describe to generate version


esword:
## What happened?
The plugin uses some complex, looping logic to calculate the describe string which forms the basis for the version (See the Git.describe method).  This was to support older versions of git (<1.8.4) that did not have all the options that git-describe needed to produce the proper version string directly.

## What did you want to happen?
 Now that the features have been git-describe for 10+ years and the codebase has shifted to only native git usage, we can replace the custom code with a single call to git describe:

git describe --tags --always --abbrev=7 --first-parent --match=[prefix]* HEAD

ash211:
@esword can this issue be closed?

esword:
Yes

PR #705, opened 2023-03-28
Issue #704: Use built in git-describe

esword:
## Before this PR
Addresses Issue #704

## After this PR

==COMMIT_MSG==
Use built in git-describe
==COMMIT_MSG==

## Possible downsides?
Will require version of git <10 years old.  If you consider that a downside.


src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -166,29 +158,14 @@ public Boolean isClean() {
...
-            }
-
-            // No tags found, so return commit hash of HEAD
-            return GitUtils.abbrevHash(runGitCmd("rev-parse", "HEAD"));
+            return runGitCmd(
+                    "describe",
+                    "--tags",
+                    "--always",
+                    "--first-parent",
+                    "--abbrev=" + GitUtils.SHA_ABBR_LENGTH,
```
pkoenig10: We can remove `GitUtils` now

```suggestion
                    "--abbrev=7",
```

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -56,7 +57,7 @@ private boolean isClean() {
     private String description() {
         String rawDescription = expensiveComputeRawDescription();
```
pkoenig10: nit: Not this PR, but we can inline `expensiveComputeRawDescription`

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -56,7 +57,7 @@ private boolean isClean() {
...
         String processedDescription =
-                rawDescription == null ? null : rawDescription.replaceFirst("^" + args.getPrefix(), "");
+                Strings.isNullOrEmpty(rawDescription) ? null : rawDescription.replaceFirst("^" + args.getPrefix(), "");
```
pkoenig10: nit: Would prefer if we didn't change this. Treating the empty string the same as null can result in subtle bugs.

src/main/java/com/palantir/gradle/gitversion/Git.java
CRogers, line 1: Let's remove the comment about backcompat for 1.8.4 at the top of the class.

esword: done

src/main/java/com/palantir/gradle/gitversion/Git.java
esword: done

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
esword: This one we need to change because the describe method no longer returns a null  (that is, wraps a thrown exception) when the result is an empty string.  The previous implementation erroneously threw an exception because it would try to truncate the empty string when git returned no found result.

I think it's still okay because it is at this single chokepoint and passes the null on up so that all the rest of the code that calls this still behaves the same.

We could check for empty string in Git.describe() and return a null instead, but that seems worse than this.

I did do some extra cleanup of removing the unnecessary extra method in VersionDetailsImpl - The expensiveComputeRawDescription method is no longer needed since we are not caching the resulting value anymore (from Yiwen's change a couple of months ago).

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
esword: yup - just saw that and changed it.

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
pkoenig10: > We could check for empty string in Git.describe() and return a null instead, but that seems worse than this.

I think I would prefer this, primarily because it removes ambiguity from the contract of `describe` and avoids leaking concerns across function boundaries. Null means the operation failed, non-null means the operation succeeds - callers don't have think about the empty string case.

pkoenig10: Changing the return value of `Git.describe()` to an `Optional<String>` would be an even better solution, making this contract very clear to callers.

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -27,22 +25,18 @@
...
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Mimics git describe by using rev-list to support versions of git < 1.8.4.
+ * Wraps git describe.
```
pkoenig10: nit: We can just remove this comment, it's pretty useless now

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -166,30 +160,15 @@ public Boolean isClean() {
...
-
-            // No tags found, so return commit hash of HEAD
-            return GitUtils.abbrevHash(runGitCmd("rev-parse", "HEAD"));
-        } catch (IOException | InterruptedException | RuntimeException e) {
+            return runGitCommand(
+                    "describe",
+                    "--tags",
+                    "--always",
+                    "--first-parent",
+                    "--abbrev=" + SHA_ABBR_LENGTH,
```
pkoenig10: This is only used in once place, let's just inline it and avoid the string concatenation

```suggestion
                    "--abbrev=7",
```

pkoenig10 (approved):
Thanks!

PR #717, opened 2023-06-02
Add debug logs for ./gradlew printVersion

ash211:
## Before this PR
In https://github.com/palantir/gradle-git-version/issues/65#issuecomment-1574078590 I as a user was confused about the difference between `./gradlew printVersion` and `println(gitVersion())`.  I thought `./gradlew printVersion` was printing the version that gradle-git-version produced, and not the version of its gradle project.

One of the things I did while debugging was run with `--info | grep -i git` and `--debug | grep -i git`.  I think if that had included the logs included in this PR, I would've been able to determine that the problem was a missing `version gitVersion()` line, and not a bug in my tags or git repo.

## After this PR
==COMMIT_MSG==
Add debug logs for ./gradlew printVersion
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-717.v2.yml
+++ b/changelog/@unreleased/pr-717.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add debug logs for ./gradlew printVersion
+  links:
+  - https://github.com/palantir/gradle-git-version/pull/717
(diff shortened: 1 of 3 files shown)


pkoenig10:
> I thought ./gradlew printVersion was printing the version that gradle-git-version produced, and not the version of its gradle project.

I don't understand the intention of this PR. If you're using this plugin, then these values will be the same.

ash211:
Using this plugin has two steps:

https://github.com/palantir/gradle-git-version#usage

1:
```
plugins {
    id 'com.palantir.git-version' version '<current version>'
}
```

2:
```
version gitVersion()
```

This PR is intended to help debug when you've done step 1 but not 2.

PR #799, opened 2024-05-01
Cache results of git commands

pkoenig10:
https://github.com/palantir/gradle-git-version/pull/684 introduced caching for the construction of `VersionDetails`, but not for the actual Git commands. Actually running the Git commands is the most expensive part of computing the version, so we're not deriving much benefit if we're not caching those.


src/main/java/com/palantir/gradle/gitversion/TimingVersionDetails.java
```diff
@@ -16,22 +16,65 @@
...
+    @Override
+    public String getBranchName() throws IOException {
+        try (Context context = timer.start("getBranchName")) {
+            return delegate.getBranchName();
+        }
+    }
+
+    @Override
+    public String getGitHashFull() throws IOException {
+        try (Context context = timer.start("getBranchName")) {
```
felixdesouza: timers are named wrong here

pkoenig10: Oops, fixed

felixdesouza:
This has been effectively fixed with configuration cache via #967 and #975

PR #807, opened 2024-06-04
Remove timer

pkoenig10:
The acute motivation for this PR is to avoid the deprecation warnings when modifying the legacy `gradleEnterprise`.`

```
WARNING: The following functionality has been deprecated and will be removed in the next major release of the Develocity Gradle plugin. Run with '-Ddevelocity.deprecation.captureOrigin=true' to see where the deprecated functionality is being used. For assistance with migration, see https://gradle.com/help/gradle-plugin-develocity-migration.
- The deprecated "gradleEnterprise.buildScan.buildFinished" API has been replaced by "develocity.buildScan.buildFinished"
- The deprecated "gradleEnterprise.buildScan.value" API has been replaced by "develocity.buildScan.value"
```

My original intention was simply to modify `BuildScanPluginInterop` to also support the `develocity` plugin. However, while testing I noticed that `BuildScanPluginInterop` is simply broken. The reported value is always zero because the code that runs the tasks uses a different `GitVersionCacheService` instance than the one used to record the custom values.

I think this has been broken for quite a while. Given that no one has noticed, it is clearly not that important. Thus, it seemed better to just remove the code rather than "fix" code that was just broken anyways.


felixdesouza:
👍  I think that's fair. I think for the most part, a lot of this can go away once we use configuration cache safe providers. Then we don't have to worry about build service's or random caches ourselves as gradle should do it for us once at the start of the build.

Thanks for putting this together.

PR #875, opened 2024-12-09
groovy dependencies no longer needed

esword:
## Before this PR
the groovy dependencies are no longer needed - the performance code was ripped out long ago.  See #807 

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


bjlaub:
👍

PR #928, opened 2025-03-25
Add GIT_VERSION env var support

FinlayRJW:
## Before this PR
Previously the [gradle guide](https://github.com/palantir/gradle-guide/blob/45c8c702b3c712ede5aa869f078752434d80356d/guide/testing-and-running-your-plugins-locally.md#running-your-plugin-in-other-repos-locally) required users to have this line to there `build.gradle`

```gradle
version System.env.CIRCLE_TAG ?: gitVersion()
```

## After this PR
Now `gradle-git-version` checks for a new ENV VAR `GIT_VERSION` that acts as an override so that users can publish at different versions from default e.g.

`GIT_VERSION=999 ./gradlew pTML`

Without requiring the `CIRCLE_TAG` workaround

==COMMIT_MSG==
Add GIT_VERSION env var support
==COMMIT_MSG==

## Possible downsides?
If this is rolled out we will now use `gitVersion()` all the time and not use `CIRCLE_TAG` this should be okay as I think `CIRCLE_TAG` matches the git version

### DIFF
--- a/changelog/@unreleased/pr-928.v2.yml
+++ b/changelog/@unreleased/pr-928.v2.yml
@@ -1,5 +1,5 @@
 type: improvement
 improvement:
-  description: Add GIT_VERSION env var support
+  description: Add `GIT_VERSION` env var support
   links:
   - https://github.com/palantir/gradle-git-version/pull/928
--- a/changelog/@unreleased/pr-928.v2.yml
+++ b/changelog/@unreleased/pr-928.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add GIT_VERSION env var support
+  links:
+  - https://github.com/palantir/gradle-git-version/pull/928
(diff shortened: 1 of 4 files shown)


changelog/@unreleased/pr-928.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add GIT_VERSION env var support
```
felixdesouza:
```suggestion
  description: Add `GIT_VERSION` env var support
```

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -193,6 +193,28 @@ class GitVersionPluginTests extends Specification {
...
+            version gitVersion()
+        '''.stripIndent()
+        gitIgnoreFile << 'build'
+        Git git = new Git(projectDir, true)
+        git.runGitCommand("init", projectDir.toString())
+        git.runGitCommand("add", ".")
+        git.runGitCommand("commit", "-m", "'initial commit'")
+        git.runGitCommand("tag", "-a", "1.0.0", "-m", "1.0.0")
+
+        when:
```
felixdesouza, line 211: nit: let's just ensure that the version without 999 is what you want i.e. 1.0.0

felixdesouza (approved):

PR #962, opened 2025-05-19
Allow numbers in prefixes

davidpaquin:
## Before this PR

Prefixes were restricted to `[/@]?([A-Za-z]+[/@-])+`, which meant numbers were not allowed. Tags whose product names included numbers, e.g. `@foo/my-product-v2@` or `@a2z/my-product` were not compatible, even though there is no risk to allowing them.

When the feature was added, there was discussion about scoping down the valid prefix space, but the conclusion was that numbers should be fine ([link](https://github.com/palantir/gradle-git-version/pull/61#discussion_r353129294)).

## After this PR

==COMMIT_MSG==
Allow numbers in prefixes.
==COMMIT_MSG==

## Possible downsides?

There's probably an argument to not allow prefixes to start with a number, but because the prefix must end with a delimiter, I don't think it's necessary to enforce that.

### DIFF
--- a/changelog/@unreleased/pr-962.v2.yml
+++ b/changelog/@unreleased/pr-962.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Allow numbers in prefixes.
+  links:
+  - https://github.com/palantir/gradle-git-version/pull/962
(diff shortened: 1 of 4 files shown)


pkoenig10:
Can you please update the PR description to include the motivation for this PR? What are you trying to accomplish?

esword (commented):
please update the note in the readme about valid pattern

esword (approved):

PR #967, opened 2025-06-03
Make `gradle-git-version` configuration-cache compatible

FinlayRJW:
## Before this PR

gradle-git-version is not configuration cacheable, due to the following problems:

Screenshot 2025-06-03 at 3 46 34 PM

## After this PR

Make gradle-git-version configuration-cacheable. Add an integration test to prevent regressions. 

==COMMIT_MSG==
Make gradle-git-version configuration-cacheable
==COMMIT_MSG==


src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -55,32 +52,16 @@ private String runGitCmd(String... commands) throws IOException, InterruptedExce
...
-            builder.append(System.getProperty("line.separator"));
-        }
-
-        int exitCode = process.waitFor();
-        if (exitCode != 0) {
-            return "";
-        }
-
-        return builder.toString().trim();
+        return providerFactory
```
FinlayRJW: might want to check exit code

build.gradle
```diff
@@ -52,6 +52,7 @@ dependencies {
     compileOnly 'org.immutables:value::annotations'
 
     testImplementation gradleTestKit()
+    testImplementation 'com.netflix.nebula:nebula-test' // for better temp directory junit rule only
```
FinlayRJW: nit: what does the comment `// for better temp directory junit rule only` refer to?

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -17,34 +17,32 @@
...
     private final File directory;
+    private final ProviderFactory providerFactory;
 
-    Git(File directory) {
-        this(directory, false);
+    Git(File directory, ProviderFactory providerFactory) {
+        this(directory, false, providerFactory);
     }
 
     @VisibleForTesting
```
FinlayRJW: I suspect you dont need this to be visible for testing anymore

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -17,34 +17,32 @@
...
-    Git(File directory, boolean testing) {
+    Git(File directory, boolean testing, ProviderFactory providerFactory) {
+        this.providerFactory = providerFactory;
+        this.directory = directory;
         if (!gitCommandExists()) {
             throw new RuntimeException("Git not found in project");
         }
-        this.directory = directory;
         if (testing && !checkIfUserIsSet()) {
```
FinlayRJW: don't need `setGitUser` or `checkIfUserIsSet` anymore

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,45 @@
...
+import nebula.test.IntegrationTestKitSpec
+
+class ConfigurationCacheTest extends IntegrationTestKitSpec {
+
+    def setup() {
+        // language=Gradle
+        buildFile << """
+            apply plugin: 'com.palantir.git-version'
+            apply plugin: 'java-library'
+
```
FinlayRJW, line 33: this test wont prevent regressions currently as the git versions plugin is not doing anything

you need to add `version gitVersion()` in here and that should cause the plugin to be configured

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,45 @@
...
+                mavenCentral()
+                mavenLocal()
+            }
+        """.stripIndent(true)
+
+        definePluginOutsideOfPluginBlock = true
+        keepFiles = true
+    }
+
+    def "help task runs with configuration cache"() {
```
FinlayRJW: test name does not match what test does

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,45 @@
...
+
+        expect:
+        createRunner('classes', '--dry-run', '--configuration-cache')
```
FinlayRJW: try bumping this to build - you may have fixed it for that as well so better to set at a higher level if you can

build.gradle
kelvinou01: I copy pasted it from another repo. Doesn't seem significant to me so I'll remove it.

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
kelvinou01: yep it works for build as well!

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
kelvinou01: good catch!

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
kelvinou01: Got it. gitVersion is a closure brought in via an extension, not a task. So we have to call it to test for configuration-phase regressions.

src/main/java/com/palantir/gradle/gitversion/Git.java
kelvinou01: Yep good catch!

changelog/@unreleased/pr-967.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Configuration cache
```
felixdesouza: This is what goes in the releases page, let's make it clear what it's actually doing. https://github.com/palantir/gradle-git-version/releases/tag/3.3.0 is a good example, https://github.com/palantir/gradle-git-version/releases/tag/3.2.0 not so much.

```suggestion
  description: Make `gradle-git-version` configuration cache compatible.
```

src/test/java/com/palantir/gradle/gitversion/VersionDetailsTest.java
```diff
@@ -24,11 +24,13 @@
 import java.nio.file.Files;
 import java.util.HashMap;
 import java.util.Map;
+import javax.inject.Inject;
+import org.gradle.api.provider.ProviderFactory;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
-public class VersionDetailsTest {
```
felixdesouza: By making this abstract, these tests no longer run. Note this is just normal junit tests, it has no gradle harnesses. If it were running, the `@Inject` wouldn't do anything. In fact the test *couldn't* run unless it knew what to do with the `@Inject`. You can't get an instance of this class without providing an implementation for `getProviderFactory`

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -750,4 +749,50 @@ class GitVersionPluginTests extends Specification {
...
+        if (exitCode != 0) {
+            return "";
+        }
+
+        return builder.toString().trim();
+    };
+
+    private static void gitInit(File projectDir) {
+        if (runGitCmd(projectDir, [:], "config", "user.email").isEmpty()) {
+            // For some reason, the tests need --global to pass on CI, but not local
```
felixdesouza: you will always have a `user.email` set on your local machine, so something else is going wrong there I think

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -750,4 +749,50 @@ class GitVersionPluginTests extends Specification {
...
+        int exitCode = process.waitFor();
+        if (exitCode != 0) {
+            return "";
+        }
+
+        return builder.toString().trim();
+    };
+
+    private static void gitInit(File projectDir) {
+        if (runGitCmd(projectDir, [:], "config", "user.email").isEmpty()) {
```
felixdesouza: you have to do this command *after* you `git init`, otherwise you're just in a random folder, deep with `gradle-git-version's` git folder.

Once you do the git init, then you can start setting the user email and name. I also don't think you need to use `--global` as that will affect system state, which isn't amazing. It should just be fine without the flag or just `--local`

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -750,4 +749,50 @@ class GitVersionPluginTests extends Specification {
...
+            runGitCmd(projectDir, [:], "config", "--global", "user.email", "email@example.com")
+            runGitCmd(projectDir, [:], "config", "--global", "user.name", "name")
+        }
+        runGitCmd(projectDir, [:], "init", projectDir.toString())
+
+        // So git doesn't ask you to gpgsign when running tests locally
+        runGitCmd(projectDir, [:], "config", "commit.gpgsign", "false")
+        runGitCmd(projectDir, [:], "config", "tag.gpgsign", "false")
+        runGitCmd(projectDir, [:], "config", "tag.forcesignannotated", "false")
+
```
felixdesouza:
```suggestion
        runGitCmd(projectDir, [:], "init", projectDir.toString())

        // So git doesn't ask you to gpgsign when running tests locally
        runGitCmd(projectDir, [:], "config", "commit.gpgsign", "false")
        runGitCmd(projectDir, [:], "config", "tag.gpgsign", "false")
        runGitCmd(projectDir, [:], "config", "tag.forcesignannotated", "false")
        runGitCmd(projectDir, [:], "config", "user.email", "email@example.com")
        runGitCmd(projectDir, [:], "config", "user.name", "name")
```

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,46 @@
...
+
+        definePluginOutsideOfPluginBlock = true
+        keepFiles = true
+    }
+
+    def "build task runs with configuration cache"() {
+        expect:
+        createRunner('build', '--dry-run', '--configuration-cache')
+                .build()
+    }
```
felixdesouza, line 89: I'm not sure what the best way to test this is as I haven't looked in the details, but it should also be the case that we figure out whether or not hte configuration cache was used a second time round, which is the eventual goal right

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,46 @@
...
+        buildFile << """
+            apply plugin: 'com.palantir.git-version'
+            apply plugin: 'java-library'
+
+            repositories {
+                mavenCentral()
+                mavenLocal()
+            }
+
+            version gitVersion()
```
felixdesouza, line 39: if we have to exercise the configuration cache inputs via these calls, then we should also do it for the other apis `gradle-git-version` offers.

That would likely mean move this out of `setup` and into specific test methods for each piece we offer

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
```diff
@@ -750,4 +749,50 @@ class GitVersionPluginTests extends Specification {
 
         return gradleRunner
     }
+
+
+    private static String runGitCmd(File directory, Map<String, String> envvars, String... commands)
```
felixdesouza, line 757: the envvars map is empty most of the time, can we just get an overload:

```groovy
runGitCmd(File directory, String... commands) {
  runGitCmd(directory, [:], commands)
}
```

build.gradle
felixdesouza: I'm not really understanding where this is used?

changelog/@unreleased/pr-967.v2.yml
kelvinou01: Gotcha!

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -18,21 +18,26 @@
...
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import javax.inject.Inject;
 import org.gradle.api.Project;
 import org.gradle.api.provider.Provider;
+import org.gradle.api.provider.ProviderFactory;
 import org.gradle.api.services.BuildService;
 import org.gradle.api.services.BuildServiceParameters;
 
 public abstract class GitVersionCacheService implements BuildService<BuildServiceParameters.None> {
```
felixdesouza, line 28: This class doesn't really make a lot of sense, it appears to just reduce object allocations, but the calculation is never cached, it's always ran again.

I did a couple of tests regarding how many times things were run:
* if you create a provider with the same args, multiple times, they will be invoked multiple times, the output will be saved for and deduplicated for configuration cache usage. i.e. if you do a restore, the same value will be used for all instances of the said provider.
* if you create a provider with the same args, but share it everywhere, then it will only be invoked once. So like a memoized supplier.

So if we wanted to reduce duplicate calls, we'd actually need to stash a provider inside `VersionDetailsImpl`, and then calls to methods in `VersionDetailsImpl` will get that value out of said provider, as currently we are making new providers for each call.

This should probably be a separate PR seeing as it's strictly an improvement on the status quo, and it makes the the non-cc mode where you've invalidated the cache faster too.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
```diff
@@ -18,21 +18,26 @@
...
 
     private final ConcurrentMap<String, VersionDetails> versionDetailsMap = new ConcurrentHashMap<>();
 
+    @Inject
+    protected abstract ProviderFactory getProviderFactory();
```
felixdesouza, line 33: if you also inject an `ObjectFactory`, then you can avoid passing a `ProviderFactory` into `VersionDetailsImpl` and instead just call [`ObjectFactory#newInstance`](https://docs.gradle.org/current/javadoc/org/gradle/api/model/ObjectFactory.html#newInstance(java.lang.Class,java.lang.Object...))

```java
ObjectFactory.newInstance(VersionDetailsImpl.class, getDir, gitVersionArgs)
```

inside `VersionDetailsImpl.class`, you'd also inject `ProviderFactory`.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
kelvinou01: This seems like a separate issue to me. Let's fix it in a separate PR!

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
kelvinou01: Not sure how that slipped past me......I've fixed it now!

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
kelvinou01: Good idea! Let's test for this for all the `ConfigurationCacheTest`s we write for the other plugins as well.

src/test/groovy/com/palantir/gradle/gitversion/GitVersionPluginTests.groovy
kelvinou01: Good idea!

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
kelvinou01: Also, as we've discussed privately, --dry-run doesn't catch tasks that aren't written in a configuration cache friendly way. Let's remove --dry-run.

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
kelvinou01: `VersionDetailsImpl` is `final`, and can't be `abstract`. This means injecting isn't possible. 

Is there a good reason why this class is `final`? If not, let's make it `abstract`?

src/test/java/com/palantir/gradle/gitversion/VersionDetailsTest.java
kelvinou01: Thanks, and good catch. I'll remember this for future PRs.

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
kelvinou01: Alright, I've created two test cases for the two methods we offer via this plugin. Does it look good to you?

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
felixdesouza: yeah, make it abstract

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,87 @@
...
+        setupBuildWithGitVersion("""
+            apply plugin: 'com.palantir.git-version'
+            apply plugin: 'java-library'
+
+            repositories {
+                mavenCentral()
+                mavenLocal()
+            }
+
+            versionDetails versionDetails()
```
felixdesouza: I think you need to call methods on `versionDetails` ? As if you remember, it doesn't do anything by itself, and you need to invoke the git commands on them

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,87 @@
...
+    private void setupBuildWithGitVersion(String buildFileContent) {
+        // language=Gradle
+        buildFile << buildFileContent.stripIndent(true)
+    }
+
+    /**
+     * Runs the specified tasks twice with configuration cache and verifies cache behavior.
+     * Returns true if the configuration cache was properly used on the second run.
+     */
+    private boolean runTasksWithConfigurationCache(String... tasks) {
```
felixdesouza, line 79: nice

src/test/java/com/palantir/gradle/gitversion/VersionDetailsTest.java
```diff
@@ -79,14 +120,16 @@ public void short_sha_when_no_annotated_tags_are_present() {
                 "--date=" + formattedTime,
                 "--allow-empty");
 
-        assertThat(versionDetails().getVersion()).isEqualTo("f0f4555");
+        assertThat(versionDetails().getVersion()).isEqualTo("f2bc772");
```
felixdesouza, line 123: what happened here?

src/test/java/com/palantir/gradle/gitversion/VersionDetailsTest.java
```diff
@@ -33,14 +35,52 @@ public class VersionDetailsTest {
...
+        git.runGitCommand("config", "tag.gpgsign", "false");
+        git.runGitCommand("config", "tag.forcesignannotated", "false");
+        git.runGitCommand("config", "user.email", "email@example.com");
+        git.runGitCommand("config", "user.name", "name");
+    }
+
+    private void createGitIgnore(File dir) {
+        File gitignoreFile = new File(dir, ".gitignore");
+
+        String gitignoreContent = "# Gradle project files\n" + ".gradle/\n"
```
felixdesouza, line 59: just use multiline strings?

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
felixdesouza: under the hood, for `gitVersion`, it gets a version details, and *then* it calls version underneath that.

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
```diff
@@ -0,0 +1,87 @@
...
+        def firstRun = createRunner(tasks + ['--configuration-cache'] as String[]).build()
```
felixdesouza, line 80: does this ensure that the tasks have succeeded?

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheService.java
felixdesouza: for future reference, anything that requires gradle to instantiate will require it to be abstract, so assume that's what I/anyone else will mean

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
kelvinou01: Yep, versionDetails doesn't do anything by itself. Thanks for the tip.

src/test/java/com/palantir/gradle/gitversion/VersionDetailsTest.java
kelvinou01: We added a .gitignore, because we need to ignore the files `this.project` brings in. We added `this.project` because `Git` was made project-aware, so gradle could be made aware of external calls for configuration-cache compatibility.

src/test/java/com/palantir/gradle/gitversion/VersionDetailsTest.java
kelvinou01: I think that's a java 15 feature, and [we're on java 8](https://github.com/palantir/gradle-git-version/blob/6732946904792be4253c4fb416a874bd9d86cb0c/build.gradle#L63)

src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy
kelvinou01: Yes, this fails if e.g. a problem was found with storing the configuration cache, or if one of the tasks failed.

src/test/java/com/palantir/gradle/gitversion/VersionDetailsTest.java
felixdesouza: yeah, let's get this in, and let's have a FLUP to bump this plugin to java 17. Major releasing when you do so.

felixdesouza (approved):
nice! 

FLUPs:
* avoid doing multiple calls, for the same set of args (this should be a relatively small refactor)
* bump this repo to be java 17 and use baseline-java-versions

PR #972, opened 2025-06-10
Increasing minimum supported JDK to JDK 17

crogoz:
## Before this PR

Minimum JDK version supported is JDK 8.

## After this PR

internal slack thread [here](https://pl.ntr/2ur). 
In preparation for Gradle 9 which will require a minimum Java 17 version.
- Use `baseline-java-versions`
- Sets the sourceCompatibility to JDK 17

==COMMIT_MSG==
Increasing minimum supported JDK to JDK 17
==COMMIT_MSG==


build.gradle
```diff
@@ -14,22 +14,25 @@ buildscript {
...
 }
 
 plugins {
     id 'com.gradle.plugin-publish' version '1.3.1'
-id 'com.palantir.git-version' version '3.3.0'
+    id 'com.palantir.git-version' version '3.3.0'
     id 'com.palantir.idea-test-fix' version '0.1.0'
- id 'com.palantir.baseline' version '6.32.0'
-id 'com.palantir.consistent-versions' version '2.34.0'
+    id 'com.palantir.baseline' version '6.32.0'
```
felixdesouza: do we want both this and lines 17? they're conflicting versions no?

felixdesouza: okay, ran it through a build scan, it doesn't actually matter, it will just use `6.32.0`. But it is confusing, can we pick which one we want to use? i.e. either `apply plugin: 'com.palantir.baseline'` or apply it here in the plugins block.

Not sure how it works with `baseline-java-versions`, whether it can be applied via the plugins block too.

felixdesouza (commented):
Can we also update the README to say that this is a JDK 17+ PR in preparation for Gradle 9.

Or maybe we can drop the Gradle 9 bit, but just that we're JDK 17.

build.gradle
crogoz: thanks for noticing! I just tried to indent this block without reading it through. addressed in https://github.com/palantir/gradle-git-version/pull/972/commits/c3426bbff705921100a9da566ff4b6213735184e

felixdesouza: I think you'll need to bump the version in the `classpath` declaration

felixdesouza (dismissed):

felixdesouza:
👍

PR #975, opened 2025-06-13
Cache calls to git in `VersionDetailsImpl`

kelvinou01:
## Before this PR

[FLUP to 967](https://github.com/palantir/gradle-git-version/pull/967#pullrequestreview-2904436620)

We already have `GitVersionCacheService` as a `BuildService`. However, the `VersionDetails` objects it returns does no cacheing — repeated calls to e.g. `VersionDetails::describe` do the same repeated calls to `git`

Also, `VersionDetailsTest::git_version_result_is_being_cached` never fails, even without cacheing (e.g. using two separate instances of `VersionDetailsImpl`). This is because the two tags 1.00 and 2.00 are pointing towards the same commit. `git describe` returns the oldest tag of that commit, which is `1.0.0`. 

## After this PR

This PR simplifies the interface of `Git` to keep it on one level of abstraction, implements cacheing in `VersionDetailsTest`, and fixes `VersionDetailsTest::git_version_result_is_being_cached` to test cacheing behavior. 

Other than cacheing,. I kept the behavior of VersionDetails the same, (sometimes returning `null` upon errors) to preserve public facing behavior.

==COMMIT_MSG==
Cache calls to git in `VersionDetailsImpl`
==COMMIT_MSG==

## Possible downsides?

`VersionDetails` now caches calls to git. This workflow breaks correctness: do git commit and tag ==> `getVersionDetails` and do stuff with it ==>  do commit and tag ==> do more stuff with the same `VersionDetails`.


src/main/java/com/palantir/gradle/gitversion/GradleAwareGit.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Runs git commands in a gradle-aware manner, so gradle can do up-to-date checking for the configuration cache.
+ */
+class GradleAwareGit {
+    public static class GitCommandFailed extends Exception {
+        @SuppressWarnings("checkstyle:visibilitymodifier")
+        public int exitCode;
+
+        @SuppressWarnings("checkstyle:visibilitymodifier")
+        public String errorOutput;
```
felixdesouza: if you have to, make this a record, we typically do not expose internal state of a class, even for simple classes. The checkstyle violation is not spurious

src/main/java/com/palantir/gradle/gitversion/GradleAwareGit.java
```diff
@@ -0,0 +1,94 @@
...
+                            execSpec.args("version");
+                            execSpec.workingDir(repositoryDir);
+                        })
+                        .getResult()
+                        .get()
+                        .getExitValue()
+                == 0;
+    }
+
+    public String run(Map<String, String> envvars, String... commands) throws GitCommandFailed {
```
felixdesouza: These should return the `Provider` otherwise it's just realising the provider here, and there's no way to use the caching of this `ProviderFactory`. 

Or better still, it should return some sort of type: `GitExecOutput`, which looks very similar to the `ExecOutput`:

```java
interface GitExecOutput {
   Provider<String> getGitOutput(); // do your trimming here and handling here
}
```

you'd probably make this an abstract class though however.

you'd probably need to do some sort of `zip/flatMap/orElse` with all three bits of three bits of data from `ExecOutput`.

src/main/java/com/palantir/gradle/gitversion/GradleAwareGit.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(GradleAwareGit.class);
+
+    private final File repositoryDir;
+    private final ProviderFactory providerFactory;
+
+    GradleAwareGit(File repositoryDir, ProviderFactory providerFactory) {
+        this.providerFactory = providerFactory;
+        this.repositoryDir = repositoryDir;
+        if (!gitCommandExists()) {
```
felixdesouza: I don't think this should be here, every version details you make, makes one of these, and the result is not cached. You should pull this into something higher up. Perhaps in the cache service. If that's as a field somewhere, it would be invoked once and cached.

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -17,29 +17,43 @@
...
     private static final Logger log = LoggerFactory.getLogger(VersionDetailsImpl.class);
     private static final int VERSION_ABBR_LENGTH = 10;
 
     private static final String DOT_GIT_DIR_PATH = "/.git";
-    private final GitVersionArgs args;
+    private final GitVersionArgs gitVersionArgs;
 
-    private Git nativeGitInvoker;
+    private GradleAwareGit gradleAwareGit;
+    private final Map<String, String> gitCache = new ConcurrentHashMap<>();
```
felixdesouza: so you needn't do all of this. For each of the methods offered up in `VersionDetails`, they should be a provider. Gradle will cache the result for you, and you don't need to maintain the cache. i.e.

```java
class VersionDetailsImpl implements VersionDetails {
  private final Provider<Boolean> isCleanProvider;
  // rest of them

  VersionDetailsImpl(ProviderFactory providerFactory, File gitDir, GitVersionArgs gitVersionArgs) {
    // usual constructor args
    isCleanProvider = gradleAwareGit.run("status", "--porcelain").map(output -> output.isEmpty());
    // all the rest of the things we're offering.
  }

}
```

I actually think you want an additional class, `LazyVersionDetails` which offers `Provider<T>` of all the methods in `VersionDetails`. `VersionDetails` takes a `LazyVersionDetails` in, and then just calls get on their respective methods.

src/test/java/com/palantir/gradle/gitversion/VersionDetailsTest.java
```diff
@@ -36,21 +37,25 @@ public class VersionDetailsTest {
     public File temporaryFolder;
 
     private Project project;
-    private Git git;
+    private GradleAwareGit gradleAwareGit;
 
     final String formattedTime = "'2005-04-07T22:13:13'";
 
     @BeforeEach
     public void before() {
```
felixdesouza: just add a `throws Exception` here.

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -17,29 +17,43 @@
...
+        this.gitVersionArgs = gitVersionArgs;
+    }
+
+    private String runGitWithCacheing(String... args) throws GitCommandFailed {
+        String key = Arrays.toString(args);
+
+        if (!gitCache.containsKey(key)) {
+            gitCache.put(key, gradleAwareGit.run(args));
+        }
+        return gitCache.get(key);
```
felixdesouza: not that it matters, because we shouldn't be doing the caching ourselves, but I would have just passed in a separate key, instead of doing `Array.toString`, the key should just be name of the method that's calling `runGitWithCaching`.

e.g. 

```java
private boolean isClean() { 
  return runGitWithCaching("isClean", "status", "--porcelain")
}
```

src/main/java/com/palantir/gradle/gitversion/GradleAwareGit.java
felixdesouza: Judging from other comment, I don't think you need this class tbh

felixdesouza: Just throw a `RuntimeException` inside the provider you're returning when it doesn't work out.

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -17,29 +17,43 @@
...
+    private GradleAwareGit gradleAwareGit;
+    private final Map<String, String> gitCache = new ConcurrentHashMap<>();
 
-    VersionDetailsImpl(ProviderFactory providerFactory, File gitDir, GitVersionArgs args) {
+    VersionDetailsImpl(ProviderFactory providerFactory, File gitDir, GitVersionArgs gitVersionArgs) {
         String gitDirStr = gitDir.toString();
         String projectDir = gitDirStr.substring(0, gitDirStr.length() - DOT_GIT_DIR_PATH.length());
-        this.nativeGitInvoker = new Git(new File(projectDir), providerFactory);
-        this.args = args;
+        this.gradleAwareGit = new GradleAwareGit(new File(projectDir), providerFactory);
```
felixdesouza: perhaps we can just pass in the `GradleAwareGit` into this class. Rather than pass its dependencies and then construct it. This makes it easy to do what I was referring to with `gitCommandExists`.

src/main/java/com/palantir/gradle/gitversion/GradleAwareGit.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Map;
+import org.gradle.api.provider.ProviderFactory;
+import org.gradle.process.ExecOutput;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Runs git commands in a gradle-aware manner, so gradle can do up-to-date checking for the configuration cache.
+ */
+class GradleAwareGit {
```
felixdesouza: I don't think it's worth the rename, it seems to suggest there is a non-gradle aware git, which isn't actually the case. It's an internal type, and I think should be fine to keep as `Git`. Or `GitInvoker` idk.

felixdesouza (commented):
From `ProviderFactory` javadoc:

> Allows lazy access to the output of the external process.
When the process output is read at configuration time it is considered as an input to the configuration model. Consequent builds will re-execute the process to obtain the output and check if the cached model is still up-to-date.

You don't have to maintain your own cache!

felixdesouza:
> VersionDetails now caches calls to git. This workflow breaks correctness: do git commit and tag ==> getVersionDetails and do stuff with it ==> do commit and tag ==> do more stuff with the same VersionDetails.

1. This only really matters if you are trying to do `git` write calls within the same gradle session. Every gradle command is effectively a new "build" session. So when gradle runs again, you will make all your structures again.
2. Breaking correctness is a scary thing to put on a PR 😂

src/main/java/com/palantir/gradle/gitversion/GradleAwareGit.java
kelvinou01: The providerFactory here was just a way to tell gradle that this is an external call. Is there a way to utilize providers for cacheing here?

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
kelvinou01: Should we use Providers as a generic "cacheable field", or as a Gradle-specific type?

My understanding is that Provider is a Gradle abstraction for deferred and cacheable values, typically used within Gradle-managed objects (like tasks or extensions). Since VersionDetailsImpl is just a POJO, not managed by Gradle, I thought it would be more appropriate to use a simple cache (like a HashMap).

Using Provider outside of Gradle-managed contexts might introduce unnecessary coupling to Gradle APIs, and could make the code harder to test or reuse outside of Gradle.

felixdesouza: This is a gradle plugin, not a generic git library. People will be interacting with it via Gradle in some way shape or form. The plugin is old enough that `Provider`'s were not really a thing widely used. If it were written from scratch today, it would almost certainly be a type returning `Provider`'s for each thing. We can't get rid of it since it's API, but we can definitely add a separate `LazyVersionDetails` that returns `Provider`'s for each call.

This is also perfectly suited towards configuration cache, as we want to cache these calls and have Gradle manage that state for you.

src/main/java/com/palantir/gradle/gitversion/GradleAwareGit.java
felixdesouza: Yes, you need to return (or modify with `map/flatMap/orElse`) the `Provider` that the `ProviderFactory` is giving you, instead of realising it and sticking it in a map.

kelvinou01:
> > VersionDetails now caches calls to git. This workflow breaks correctness: do git commit and tag ==> getVersionDetails and do stuff with it ==> do commit and tag ==> do more stuff with the same VersionDetails.
> 
> 1. This only really matters if you are trying to do `git` write calls within the same gradle session. Every gradle command is effectively a new "build" session. So when gradle runs again, you will make all your structures again.
> 2. Breaking correctness is a scary thing to put on a PR 😂

@felixdesouza 

1. True, this cache only works within the same VersionDetails object. Each gradle session has one version details object. 
2. Hahaha true. I thought I'd rather be upfront about any downsides, rather than hiding it, just in case people stumble upon this in the future.

felixdesouza:
That said, given that you get a new object each time gradle runs, I don't believe you have correctness issues anymore?

kelvinou01:
> That said, given that you get a new object each time gradle runs, I don't believe you have correctness issues anymore?

@felixdesouza 
Correctness issues only occur if people try to do stuff with VersionDetails in a single gradle run. I don't think anyone (hopefully) is doing this, but it's still worth noting here?

felixdesouza:
> > That said, given that you get a new object each time gradle runs, I don't believe you have correctness issues anymore?
> 
> @felixdesouza Correctness issues only occur if people try to do stuff with VersionDetails in a single gradle run. I don't think anyone (hopefully) is doing this, but it's still worth noting here?

It's worth putting this in the `README.md` in addition to the PR description. i.e. a new section on Configuration cache, and what that entails for users and the pitfall you've talked about above

src/main/java/com/palantir/gradle/gitversion/GradleAwareGit.java
kelvinou01: Hmm I think this is a good check in `Git` for correctness sake, since `GitVersionCacheService` can service multiple git repositories in it's lifetime?

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
kelvinou01: Screenshot 2025-06-13 at 5 22 12 PM

Hmm, I tried the provider approach, but seems like [Gradle doesn't cache providers yet](https://github.com/gradle/gradle/issues/25550)

felixdesouza: This isn't the same thing. Wrapping it in a provider doesn't do anything, it will call `git.run(..)` every time the provider is fetched. The issue is that you're not keeping a single reference to the provider from `ProviderFactory#exec`. You lose that when you strip the provider in the `git` call. So `git#run` needs to return a `Provider` and you need to store that as the field.

so the fields of `VersionDetailsImpl` should be set by calls to `git`, not provider wrapping. It should be very similar to my above comment: https://github.com/palantir/gradle-git-version/pull/975#discussion_r2144817592

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -21,25 +21,39 @@
...
+    private Git git;
+
+    private Provider<String> description;
+    private Provider<Boolean> isClean;
+    private Provider<String> gitHashFull;
+    private Provider<String> branchName;
+
+    VersionDetailsImpl(ProviderFactory providerFactory, File gitDir, GitVersionArgs gitVersionArgs) {
+        String projectDir = gitDir.getParent();
+        this.git = new Git(new File(projectDir), providerFactory);
```
felixdesouza: don't think you need this field anymore?

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -99,24 +120,34 @@ public String getLastTag() {
...
-    public String getBranchName() throws IOException {
-        return nativeGitInvoker.getCurrentBranch();
+    public String getBranchName() {
+        try {
+            String branchName = this.branchName.get();
+            if (branchName.isEmpty()) {
+                return null;
+            } else {
+                return branchName;
+            }
```
felixdesouza: just use `Strings.emptyToNull(branchName)`. This is from guava

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -17,130 +17,65 @@
...
         if (!gitCommandExists()) {
-            throw new RuntimeException("Git not found in project");
+            throw new RuntimeException("git command not found in project.");
         }
     }
 
-    private String runGitCmd(String... commands) throws IOException, InterruptedException {
-        return runGitCmd(new HashMap<>(), commands);
+    private boolean gitCommandExists() {
+        return providerFactory
```
felixdesouza: I still don't believe this should be here. It's not even tied to this particular git repo.

You also want to ensure with `execSpec.setIgnoreExitValue(true)` otherwise you're not actually going to be able to throw the exception providing a nice helper message i.e. `"git command not found in project"`.

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -17,130 +17,65 @@
...
     private final ProviderFactory providerFactory;
 
-    Git(File directory, ProviderFactory providerFactory) {
+    Git(File repositoryDir, ProviderFactory providerFactory) {
         this.providerFactory = providerFactory;
-        this.directory = directory;
+        this.repositoryDir = repositoryDir;
         if (!gitCommandExists()) {
-            throw new RuntimeException("Git not found in project");
+            throw new RuntimeException("git command not found in project.");
```
felixdesouza: This message makes no sense, granted you didn't add it, but I don't think it makes sense to keep it. `git` isn't a project scoped file, it's a command.

src/main/java/com/palantir/gradle/gitversion/Git.java
felixdesouza: Probably not worth keeping it at all, we've not got anything similar in all the other `git` stuff we did.

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -56,14 +70,21 @@ public String getVersion() {
...
-        String processedDescription =
-                rawDescription == null ? null : rawDescription.replaceFirst("^" + args.getPrefix(), "");
-        return processedDescription;
+        try {
+            String description = this.description.get();
+            if (description.isEmpty()) {
+                return null;
+            } else {
+                return description;
+            }
```
felixdesouza: same thing here

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
kelvinou01: From our discussion:
- [Providers don't cache in general](https://github.com/palantir/gradle-git-version/pull/975#discussion_r2144817592)
- [However, external processes are cached](https://docs.gradle.org/current/javadoc/org/gradle/process/ExecOutput.html#getResult()), i.e. they only run once 
- The tricky part: If providerFactory.exec is called multiple times, the underlying external process e.g. git describe is run multiple times
- If we have a single reference to a describeProvider returned by providerFactory.exec, and all calls to describe() just calls describeProvider.get() , then we utilize ExecResult's cacheing, and git describe is only called once

src/main/java/com/palantir/gradle/gitversion/GradleAwareGit.java
kelvinou01: I still think `GradleAwareGit` better conveys the intent of the object compared to a more generic name like `Git`. That said, since I’m new to the team, I trust your judgment and am happy to go with `Git` as well.

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
kelvinou01: yep, good catch!

src/main/java/com/palantir/gradle/gitversion/Git.java
kelvinou01: Yep i see your point now. If git didn't exist, it should tell us in the error message when we try to use git anyways. I'll remove this check.

src/main/java/com/palantir/gradle/gitversion/Git.java
kelvinou01: yep, let's remove this whole check

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
kelvinou01: i see, that'd really make the code nicer here!

src/main/java/com/palantir/gradle/gitversion/GradleAwareGit.java
kelvinou01: As per our discussion below, let's remove this check

kelvinou01:
> It's worth putting this in the README.md in addition to the PR description. i.e. a new section on Configuration cache, and what that entails for users and the pitfall you've talked about above

@felixdesouza I've added a section to the README on git cacheing. I don't think this PR is directly part of the configuration cache workstream.

readme.md
```diff
@@ -98,6 +98,24 @@ Tasks
 This plugin adds a `printVersion` task, which will echo the project's configured version
 to standard-out.
 
+
+Cacheing
```
felixdesouza:
```suggestion
Caching
```

readme.md
```diff
@@ -98,6 +98,24 @@ Tasks
...
+---
+
+`VersionDetails` caches calls to git with gradle's [Provider\<ExecOutput\>](https://docs.gradle.org/current/javadoc/org/gradle/process/ExecOutput.html#getResult()). External calls are expensive, and cacheing prevents running the same `git` command more than once in a single build.
```
felixdesouza:
```suggestion
`VersionDetails` caches calls to `git` with Gradle's [`Provider<ExecOutput>`](https://docs.gradle.org/current/javadoc/org/gradle/process/ExecOutput.html#getResult()). External calls can be  expensive, especially if called repeatedly across multiple projects. Caching prevents running the same `git` command more than once in a single build. 
```

readme.md
```diff
@@ -98,6 +98,24 @@ Tasks
 This plugin adds a `printVersion` task, which will echo the project's configured version
 to standard-out.
 
+
+Cacheing
+---
+
+`VersionDetails` caches calls to git with gradle's [Provider\<ExecOutput\>](https://docs.gradle.org/current/javadoc/org/gradle/process/ExecOutput.html#getResult()). External calls are expensive, and cacheing prevents running the same `git` command more than once in a single build. 
+
+If the state of the git repo changes within a single gradle session (e.g. a task in the build does `git commit`), `VersionDetails` might reflect the outdated state of your repo.
```
felixdesouza:
```suggestion
> [!WARNING]
> If the state of the `git` repository changes within a single Gradle session (e.g. a task in the build does `git commit`), `VersionDetails` might reflect the outdated state of your repo.
```

readme.md
```diff
@@ -98,6 +98,24 @@ Tasks
...
+
+However, changes to the state of your repo between gradle builds i.e. two runs of gradle...
+```bash
+./gradlew check
+git commit -m "new commit"
+./gradlew check
+```
+
+...are not affected by this, as the cache is scoped to individual `VersionDetails` instances.
```
felixdesouza, line 120: Don't think we need this section

felixdesouza:
> > It's worth putting this in the README.md in addition to the PR description. i.e. a new section on Configuration cache, and what that entails for users and the pitfall you've talked about above
> 
> @felixdesouza I've added a section to the README on git cacheing. I don't think this PR is directly part of the configuration cache workstream.

It kinda is though? The `ProviderFactory#exec` api, was specifically introduced for the configuration cache. It also gives a backdrop as to why Gradle would even be caching things in the first place.

readme.md
kelvinou01: It's to reassure people who are not 100% familiar with how cacheing works in gradle. I think it’s helpful to overcommunicate here, just to make sure everyone feels confident that their use case is supported. Should I remove it?

readme.md
kelvinou01: Nice!

kelvinou01:
> > > It's worth putting this in the README.md in addition to the PR description. i.e. a new section on Configuration cache, and what that entails for users and the pitfall you've talked about above
> > 
> > 
> > @felixdesouza I've added a section to the README on git cacheing. I don't think this PR is directly part of the configuration cache workstream.
> 
> It kinda is though? The `ProviderFactory#exec` api, was specifically introduced for the configuration cache. It also gives a backdrop as to why Gradle would even be caching things in the first place.

@felixdesouza Huh, I thought it was primarily a way to cache git calls for when configuration cache entries aren't used (either invalidated, or when configuration cache is turned off). 

What configuration cache specific information should I include in the readme then? Would appreciate your guidance on this. 

Edit: added the Configuration Cache section in README after private discussion

readme.md
```diff
@@ -98,6 +98,28 @@ Tasks
 This plugin adds a `printVersion` task, which will echo the project's configured version
 to standard-out.
 
+
+☑️ Configuration cache
+---
+External process calls to `git` in this plugin are compatible with Gradle's [Configuration Cache](https://docs.gradle.org/current/userguide/configuration_cache.html).
+
+Since we're using CC compatible APIs, we leverage Gradle's ability to reuse outputs of `git` commands upon successive invocations.
```
felixdesouza:
```suggestion
Since we're using Configuration Cache compatible APIs, we leverage Gradle's ability to reuse outputs of `git` commands upon successive invocations.
```

felixdesouza (approved):

PR #1019, opened 2025-07-28
Use `com.palantir.gradle.plugintesting.ConfigurationCacheSpec`

FinlayRJW:
## Before this PR
Had our own way of running configuration cache tests

## After this PR
Now we use the same shared spec from `plugintesting` rather than having it re-implemented everywhere

==COMMIT_MSG==
Use `com.palantir.gradle.plugintesting.ConfigurationCacheSpec`
==COMMIT_MSG==


kelvinou01 (approved):
Nice!

PR #1156, opened 2026-01-13
Remove throws IOException from VersionDetails.getBranchName

palantir-sachin:
## Summary
Remove unnecessary `throws IOException` from `VersionDetails.getBranchName`

## Why
`VersionDetails.getBranchName` underlying methods don't throw `IOException` and this will make it easier to access the branch name without the need to catch

## Example Use Case
```
# another plugin apply() method

        Provider<VersionDetails> versionDetails = GitVersionCacheService.getSharedGitVersionCacheService(project)
                .map(cacheService -> cacheService.getVersionDetails(project.getProjectDir(), null));
        Provider<String> originUrl = versionDetails.map(VersionDetails::getOriginUrl);
        Provider<String> branch = versionDetails.map(VersionDetails::getBranchName);
```
==COMMIT_MSG==
Remove throws IOException from VersionDetails.getBranchName
==COMMIT_MSG==

### DIFF
--- a/src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
+++ b/src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
@@ -17,7 +17,6 @@
 package com.palantir.gradle.gitversion;
 
 import groovy.lang.Closure;
-import javax.inject.Inject;
 import org.gradle.api.Action;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
@@ -25,13 +24,6 @@ import org.gradle.api.Task;
 import org.gradle.api.provider.Provider;
 
 public final class GitVersionPlugin implements Plugin<Project> {
-
-    private final Project pluginProject;
-
-    @Inject
-    public GitVersionPlugin(Project project) {
-        this.pluginProject = project;
-    }
 
     @Override
     public void apply(final Project project) {
@@ -66,9 +58,5 @@ public final class GitVersionPlugin implements Plugin<Project> {
             task.setGroup("Versioning");
             task.setDescription("Prints the project's configured version to standard out");
         });
-    }
-
-    public Provider<GitVersionCacheService> getCacheService() {
-        return GitVersionCacheService.getSharedGitVersionCacheService(pluginProject);
     }
 }
--- a/src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
+++ b/src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
@@ -36,6 +36,7 @@ public final class GitVersionPlugin implements Plugin<Project> {
     }
 
     @Override
+    @SuppressWarnings("HiddenField")
     public void apply(final Project project) {
         project.getRootProject().getPluginManager().apply(GitVersionRootPlugin.class);
 
--- a/src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
+++ b/src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
@@ -26,19 +26,19 @@ import org.gradle.api.provider.Provider;
 
 public final class GitVersionPlugin implements Plugin<Project> {
 
-    private Project project;
-    private Provider<GitVersionCacheService> serviceProvider;
+    private final Project pluginProject;
 
     @Inject
     public GitVersionPlugin(Project project) {
-        this.project = project;
-        this.serviceProvider = GitVersionCacheService.getSharedGitVersionCacheService(project);
+        this.pluginProject = project;
     }
 
     @Override
-    @SuppressWarnings("HiddenField")
     public void apply(final Project project) {
         project.getRootProject().getPluginManager().apply(GitVersionRootPlugin.class);
+
+        Provider<GitVersionCacheService> serviceProvider =
+                GitVersionCacheService.getSharedGitVersionCacheService(project);
 
         // intentionally not using .getExtension() here for back-compat
         project.getExtensions().getExtraProperties().set("gitVersion", new Closure<String>(this, this) {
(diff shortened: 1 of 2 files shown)


felixdesouza:
I feel like it's sufficient to just get the build service from wherever you plan to use this no?

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
```diff
@@ -25,13 +26,20 @@
 
 public final class GitVersionPlugin implements Plugin<Project> {
 
+    private Project project;
+    private Provider<GitVersionCacheService> serviceProvider;
```
felixdesouza: `final`

felixdesouza: let's call this `cacheService`

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
```diff
@@ -25,13 +26,20 @@
...
+
+    @Inject
+    public GitVersionPlugin(Project project) {
+        this.project = project;
+        this.serviceProvider = GitVersionCacheService.getSharedGitVersionCacheService(project);
+    }
+
     @Override
+    @SuppressWarnings("HiddenField")
```
felixdesouza: don't think this applies anymore?

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
```diff
@@ -59,4 +67,9 @@ public void execute(Task _task) {
             task.setDescription("Prints the project's configured version to standard out");
         });
     }
+
+    public Provider<VersionDetails> getVersionDetails() {
```
felixdesouza: I think we have to be careful here. It's not clear what will be selected when being referenced from buildscripts as `getVersionDetails` will be accessible via `versionDetails` in buildscripts (potentially).

I think the number of places where this actually works is limited since you have to get the result of the plugin application.

But now you have two methods that look very similar, but one returns a realised `VersionDetails` and the other returns a `Provider<VersionDetails>` . If I were scanning the code to see how something worked, I would not be surprised if I landed on **this** method, and got confused at what *looks* like magic, despite it actually being the closure method.

To resolve this, I think we should either:
* change the name of the method
* just get what we need by using the cache service directly instead.

felixdesouza: also this method is not the same as what you get in the closure. You're always passing `null` to the details. IIRC this method was added for the internal frontend monorepo, so it's also not clear whether this gets you want you want.

palantir-sachin: makes sense. I updated the PR to expose a `getCacheService()` method. good call on the `null` to details, now the consuming project can pass `gitVersionArgs` if they want

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
```diff
@@ -25,6 +26,13 @@
 
 public final class GitVersionPlugin implements Plugin<Project> {
 
+    private final Project pluginProject;
+
+    @Inject
+    public GitVersionPlugin(Project project) {
+        this.pluginProject = project;
+    }
+
```
felixdesouza: you can actually just inject the project as a method:

```java
@Inject
protected abstract Project getProject();
```

however that's kinda naff, try to limit project references as much as possible.

Preferably, we can just use the `Gradle` instance instead.

i.e.

```java
@Inject
protected abstract Gradle getGradle();
```

If you just make an overload in `GitVersionCacheService` you should be fine.

Separately, and won't block, but you don't really even have to retrieve this through the plugin, you can just call the contents of this method.

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
```diff
@@ -59,4 +67,8 @@ public void execute(Task _task) {
             task.setDescription("Prints the project's configured version to standard out");
         });
     }
+
```
felixdesouza: If you could have a small sanity check test where you can get the cache service and then call some commands through it, would be grand. Currently, tests pass, and we know that we haven't messed up Gradle's injection instantiation, but not much else.

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
palantir-sachin: Oh good idea. I won't touch the `GitVersionPlugin` in this PR then

src/main/java/com/palantir/gradle/gitversion/GitVersionPlugin.java
palantir-sachin: Will add `GitVersionCacheService` tests in a follow up PR

src/main/java/com/palantir/gradle/gitversion/VersionDetails.java
```diff
@@ -19,7 +19,7 @@
 import java.io.IOException;
 
 public interface VersionDetails {
-    String getBranchName() throws IOException;
+    String getBranchName();
```
felixdesouza, line 22: This is not a safe change.

`IOException` is a checked exception so it needs to be dealt with in some way.

* If the caller *also* uses `throws IOException` then they're probably fine. 
* If the caller uses `try .. catch` block, it will fail to compile so existing callers will fail to compile when this upgrades. Since checked exceptions are statically declared, it can be checked at compile time that you handle it (or use `throws`). You will get the following compile error:

```
Exception 'java.io.IOException' is never thrown in the corresponding try block
```

felixdesouza:
Since the only change now is the `IOException` change, I'll close it as it's not a safe to make as [aforementioned](https://github.com/palantir/gradle-git-version/pull/1156#discussion_r2742817283).

PR #1196, opened 2026-02-20
Provide more extensible and composable API for safely invoking `git` in gradle

felixdesouza:
## Before this PR
The existing implementation works well, in that things are cached across the build and they are configuration cache safe.

Unfortunately, it's not very extensible, and `VersionDetails` is somewhat of a sorry interface. A corollary of this is that this ends up being the resting place of all possible `git` operations in Gradle, because it handles configuration cache semantics correctly, and also handles caching across the build - for non-cc invocations.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

I think the docs changes are an excellent place to start to understand how to consume this and how it solves the aforementioned issues re extensibility and composability.

Changes
* `GitVersionCacheServiceV2` is a build service, but at a lower granularity i.e. the key is for the git command being run and a given git root directory (`.git`).
* Add `GitInvoker` that will instantiate the build service, without it having to be managed by consumers.
* Introduce a `CommonGitOperations` that everything uses. This will call `GitInvoker` -> `GitVersionCacheServiceV2` -> `ProviderFactory#exec` to ensure that we use the same instance everywhere.
* use a `LoadingCache` to better signal the intent i.e. multiple calls will block and only one will actually run the equivalent of `computeIfAbsent`. N.B we can still run `ConcurrentMap#computeIfAbsent`, because we're returning a thunk, and that does not do any real work. However, if you come across and you doubt the implementation, it's annoyingly non-trivial to go through that exercise again.
* Add ability to customise `CommonGitOperations` instantiation if need be

## Possible downsides?

* I have deleted the `GitVersionCacheService` as it is not really fit for purpose anymore. There are only a few internal usages which I think we can migrate just fine. I will major version this release.
* We still potentially pay the cost of going into traversing the directory hierarchy and trying to find a `.git` folder. This is true of the existing implementation, we can fix this up in a FLUP.
* the `GIT_VERSION` override is only fed into things that call `getVersion`, so anything else that calls `description()` will not get the override. This is already true today and wanted to keep the PR down. Ideally we expose the `Description` type perhaps? Maybe a sealed class or something. Once we've figured that out, we can probably expose the `GitVersionOverride` managed type as well in the docs and for consumption.


src/main/java/com/palantir/gradle/gitversion/CommonGitOperations.java
```diff
@@ -0,0 +1,173 @@
...
+import org.gradle.api.Action;
+import org.gradle.api.model.ObjectFactory;
+import org.gradle.api.provider.Property;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.tasks.Nested;
+
+@SuppressWarnings("MoreThanOneInjectableConstructor")
+public abstract class CommonGitOperations {
+
+    public abstract static class Default extends CommonGitOperations {
```
felixdesouza, line 36: need this because you can't instantiate `Parameters` passed in as a parameter. Want to avoid making a type that is "mutable" once created where possible.

src/main/java/com/palantir/gradle/gitversion/CommonGitOperations.java
```diff
@@ -0,0 +1,173 @@
...
+        protected abstract ObjectFactory getObjectFactory();
+
+        public final CommonGitOperations create(Action<Parameters> configurator) {
+            Parameters parameters = getObjectFactory().newInstance(Parameters.class);
+            configurator.execute(parameters);
+            return getObjectFactory().newInstance(CommonGitOperations.class, parameters);
+        }
+    }
+
+    protected abstract Property<Parameters> getParameters();
```
felixdesouza: should probably make this a field and make it `final`

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -16,50 +16,79 @@
...
     private final ProviderFactory providerFactory;
+    private final ObjectFactory objectFactory;
+    private final EnvironmentVariables environmentVariables;
 
-    Git(File repositoryDir, ProviderFactory providerFactory) {
+    Git(
+            File repositoryDir,
+            ProviderFactory providerFactory,
+            ObjectFactory objectFactory,
+            EnvironmentVariables environmentVariables) {
```
felixdesouza: can't use injection here because the `ObjectFactory` in the same scope as the `BuildServiceRegistry` does not know about `ProviderFactory` as it's too high up in the scope.

I suppose we could do a similar trick with the factory perhaps. But not really pertinent to this PR.

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -16,50 +16,79 @@
...
         this.providerFactory = providerFactory;
         this.repositoryDir = repositoryDir;
+        this.objectFactory = objectFactory;
+        this.environmentVariables = environmentVariables;
     }
 
-    public Provider<String> run(Map<String, String> envvars, String... commands) {
+    interface GitParameters {
```
felixdesouza, line 67: too many permutations of different parameters being used in tests, so just used a configurator instead. Feels more canonical gradle tbh

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -16,50 +16,79 @@
...
-                String stdErr = output.getStandardError().getAsText().get();
-                String errorMsg = String.format("git command failed: %s", stdErr);
-                log.error(errorMsg);
-                throw new RuntimeException(errorMsg);
-            }
-            return output.getStandardOutput().getAsText().map(String::trim);
-        });
+        ImmutableGitExecOutput.Builder execOutputBuilder = ImmutableGitExecOutput.builder()
+                .command(gitParameters.getCommand().get());
+        return output.getResult()
```
felixdesouza: should use `gradle-utils:exec`, I still quite like the `GitExecOutput` type better for being clearer about `stdOut` being "safe" to consume.

felixdesouza: this should be easy to `.map` from one type to the other, but we can eschew all the wrangling of the types from gradle, and just map from `ExecResultWithOutput` to `GitExecOutput`.

felixdesouza: perhaps we can incorporate some concepts, not sure yet

src/main/java/com/palantir/gradle/gitversion/GitExecOutput.java
```diff
@@ -0,0 +1,70 @@
...
+
+    abstract List<String> command();
+
+    public abstract int exitCode();
+
+    public abstract String uncheckedStandardOut();
+
+    public abstract String standardError();
+
+    public final String standardOutputOfSuccessfulCommand() {
```
felixdesouza, line 38: key distinction for me, we need to ensure that values are safe to be used and that they are checked for the exit code.

src/main/java/com/palantir/gradle/gitversion/GitInvoker.java
```diff
@@ -0,0 +1,44 @@
...
+        cacheService = buildServiceRegistry.registerIfAbsent(
+                "GitVersionCacheServiceV2", GitVersionCacheServiceV2.class, _spec -> {});
+    }
+
+    public final Provider<String> invoke(String... command) {
+        return invokeWithResult(command).map(GitExecOutput::standardOutputOfSuccessfulCommand);
+    }
+
+    public final Provider<GitExecOutput> invokeWithResult(String... command) {
+        return cacheService.get().invokeWithResult(getProjectLayout(), command);
```
felixdesouza: todo flat map this

src/main/java/com/palantir/gradle/gitversion/GitVersionCacheServiceV2.java
```diff
@@ -13,41 +13,53 @@
...
+import org.gradle.api.file.ProjectLayout;
+import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.provider.Provider;
 import org.gradle.api.provider.ProviderFactory;
 import org.gradle.api.services.BuildService;
 import org.gradle.api.services.BuildServiceParameters;
 import org.gradle.api.tasks.Nested;
 
-public abstract class GitVersionCacheService implements BuildService<BuildServiceParameters.None> {
+@SuppressWarnings("RedundantModifier")
```
felixdesouza, line 33: required because gradle needs it to be public, but checkstyle complains if you have the type package private

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -16,118 +16,50 @@
...
-import org.gradle.api.provider.Provider;
-import org.gradle.api.provider.ProviderFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 class VersionDetailsImpl implements VersionDetails {
 
     private static final Logger log = LoggerFactory.getLogger(VersionDetailsImpl.class);
-    private static final int VERSION_ABBR_LENGTH = 10;
+    static final int VERSION_ABBR_LENGTH = 10;
```
felixdesouza: todo: move this to where it's used

src/test/java/com/palantir/gradle/gitversion/GitVersionPluginTests.java
```diff
@@ -476,7 +488,7 @@ println versionDetails(prefix:"my-product@").lastTag
         GitUtils.runCommands(rootProject.path().toFile(), "commit", "-m", "'initial commit'");
         GitUtils.runCommands(rootProject.path().toFile(), "tag", "-a", "my-product@1.0.0", "-m", "my-product@1.0.0");
         GitUtils.runCommands(rootProject.path().toFile(), "commit", "-m", "'commit 2'", "--allow-empty");
-        GitUtils.runCommands(rootProject.path().toFile(), "tag", "-a", "1.0.0", "-m", "1.0.0");
+        GitUtils.runCommands(rootProject.path().toFile(), "tag", "-a", "2.0.0", "-m", "2.0.0");
```
felixdesouza, line 495: the previous test made the two outcomes indistinguishable because the tags were the same and we strip the prefix

src/test/java/com/palantir/gradle/gitversion/GitVersionPluginTests.java
```diff
@@ -715,4 +727,134 @@ version gitVersion()
...
+         *
+         * method will return:
+         * <pre>
+         *     List.of(
+         *       "git describe --tags --always --first-parent --abbrev=7 '--match=*' HEAD",
+         *       "git status --porcelain"
+         *     )
+         * </pre>
+         */
+        private List<String> extractGitCalls(ArbitraryFile gitTraceFile) {
```
felixdesouza: todo: nit: static

src/test/java/com/palantir/gradle/gitversion/VersionDetailsTest.java
```diff
@@ -24,17 +24,16 @@
...
 import org.gradle.testfixtures.ProjectBuilder;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.CleanupMode;
 import org.junit.jupiter.api.io.TempDir;
 
 public class VersionDetailsTest {
 
-    @TempDir
+    @TempDir(cleanup = CleanupMode.ON_SUCCESS)
```
felixdesouza: todo: revert?

versions.props
```diff
@@ -7,3 +7,5 @@ org.junit.platform:* = 6.0.3
 org.spockframework:spock-core = 2.1-groovy-3.0
 com.netflix.nebula:nebula-test = 10.6.2
 com.palantir.gradle.utils:environment-variables = 0.23.0
+com.github.ben-manes.caffeine:caffeine = 3.2.3
+one.util:streamex = 0.8.4
```
felixdesouza: todo: new line

src/main/java/com/palantir/gradle/gitversion/Git.java
```diff
@@ -16,50 +16,79 @@
...
+    public Provider<String> run(Action<GitParameters> configureParameters) {
+        return runWithResult(configureParameters).map(GitExecOutput::standardOutputOfSuccessfulCommand);
     }
 
     public Provider<String> run(String... command) {
-        return run(new HashMap<>(), command);
+        return run(parameters -> parameters.getCommand().addAll(command));
+    }
+
+    private Map<String, String> getGitTraceEnvironmentVariables() {
```
felixdesouza, line 101: this is really neat in that you get a file containing the trace:

```
14:44:52.614346 exec-cmd.c:145          trace: resolved executable path from Darwin stack: /usr/bin/git
14:44:52.615018 exec-cmd.c:267          trace: resolved executable dir: /usr/bin
14:44:52.616169 git.c:476               trace: built-in: git describe --tags --always --first-parent --abbrev=7 '--match=*' HEAD
14:44:52.644216 exec-cmd.c:145          trace: resolved executable path from Darwin stack: /usr/bin/git
14:44:52.644921 exec-cmd.c:267          trace: resolved executable dir: /usr/bin
14:44:52.645891 git.c:476               trace: built-in: git status --porcelain
```

felixdesouza:
there's lots of missing javadoc. I think it would be good to specify behaviour. Let's agree on the shape of things and then we can add more javadocs.

src/main/java/com/palantir/gradle/gitversion/CommonGitOperations.java
```diff
@@ -0,0 +1,173 @@
...
+                    if (execOutput.exitCode() == 0) {
+                        return execOutput.uncheckedStandardOut();
+                    }
+
+                    if (execOutput.standardError().contains("Not a valid object name HEAD")) {
+                        return null;
+                    }
+
+                    // Always fails
+                    execOutput.assertNormalExitValue();
```
CRogers, line 136: Does this produce a bad error message of the form `process blah exited with exit code X`? Might we want a better error message, including the standard err at least?

src/test/java/com/palantir/gradle/gitversion/GitVersionPluginTests.java
```diff
@@ -715,4 +727,134 @@ version gitVersion()
...
+            rootProject.buildGradle().plugins().add("com.palantir.git-version");
+            StreamEx.of(rootProject, projectA, projectB)
+                    .map(GradleProject::buildGradle)
+                    .forEach(buildGradle -> buildGradle.append(
+                            """
+                            println %s
+                            println %s
+                            println %s
+                            println %s
+                            """, versionQueryMethod, versionQueryMethod, versionQueryMethod, versionQueryMethod));
```
crogoz: I think you could have something like this: 

```
.forEach(buildGradle -> buildGradle.append(versionQueryPrinter.repeat(4)));
```

where String versionQueryPrinter = String.format("println %s\n", versionQueryMethod);

src/test/java/com/palantir/gradle/gitversion/GitVersionPluginTests.java
```diff
@@ -715,4 +727,134 @@ version gitVersion()
...
+            GitUtils.gitInit(rootProject.path().toFile());
+            GitUtils.runCommands(
+                    rootProject.path().toFile(), "remote", "add", "origin", "git@github.com:example/example.git");
+            GitUtils.runCommands(rootProject.path().toFile(), "add", ".");
+            GitUtils.runCommands(rootProject.path().toFile(), "commit", "-m", "initial commit");
+            GitUtils.runCommands(rootProject.path().toFile(), "tag", "-a", "1.0.0", "-m", "1.0.0");
+
+            gradle.withArgs(
+                            "help",
+                            "-P__TESTING=true",
```
crogoz: nit: you don't need to add `-P__TESTING`, this is done automatically by the testing framework: https://github.com/palantir/gradle-plugin-testing/blob/179bf90f7db3b410422d855cfc1f5e56bf2e4372/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java#L40

src/main/java/com/palantir/gradle/gitversion/CommonGitOperations.java
```diff
@@ -0,0 +1,173 @@
...
+                        "--first-parent",
+                        "--abbrev=7",
+                        "--match=%s*".formatted(prefixValue),
+                        "HEAD")
+                .map(execOutput -> {
+                    if (execOutput.exitCode() == 0) {
+                        return execOutput.uncheckedStandardOut();
+                    }
+
+                    if (execOutput.standardError().contains("Not a valid object name HEAD")) {
```
crogoz, line 131: why do we have a special behavior for this case ? I don't see this logic in the previous implementation

src/main/java/com/palantir/gradle/gitversion/Git.java
crogoz: this looks really cool!

src/test/java/com/palantir/gradle/gitversion/GitVersionPluginTests.java
felixdesouza: amazing ❤️

src/test/java/com/palantir/gradle/gitversion/GitVersionPluginTests.java
```diff
@@ -715,4 +727,134 @@ version gitVersion()
...
+                            "-P__TESTING_GIT_TRACE=" + gitTraceFile.path().toAbsolutePath())
```
crogoz: opened: https://github.com/palantir/gradle-plugin-testing/pull/391 which might be nicer for adding extra env vars as testing properties

src/test/java/com/palantir/gradle/gitversion/GitVersionPluginTests.java
felixdesouza: is there a nice way to not have to add `TESTING_` for the environment variables generally?

felixdesouza: you read my mind!

src/main/java/com/palantir/gradle/gitversion/CommonGitOperations.java
felixdesouza: so previously, we just caught any exception and wrote it as `null`. I only special cased because it felt clearer to say the only reason something is `unspecified` or "empty" is when we don't have "a valid object name name HEAD" and other failures should fail accordingly.

src/test/java/com/palantir/gradle/gitversion/GitVersionPluginTests.java
felixdesouza: yeah this is great

src/main/java/com/palantir/gradle/gitversion/CommonGitOperations.java
felixdesouza: yeah, this just is just falling back to the standard error handling defined in [`GitExecOutput`](https://github.com/palantir/gradle-git-version/pull/1196/changes#diff-1edcb368a726ca668e090f5d62d918a9ebe8427377fcdfd6a92035a8328c6acfR48)

src/main/java/com/palantir/gradle/gitversion/CommonGitOperations.java
```diff
@@ -0,0 +1,173 @@
...
+
+        private static final Pattern GIT_DESCRIBE_PATTERN = Pattern.compile("(.*)-([0-9]+)-g.?[0-9a-fA-F]{3,}");
+
+        public String lastTag() {
+            if (isPlainTag()) {
+                return description;
+            }
+
+            Matcher match = GIT_DESCRIBE_PATTERN.matcher(description);
+            return match.matches() ? match.group(1) : null;
```
CRogers: Even though I know we're using this in a `Provider` above, I'd still rather this was an `Optional` then used with `.orElse(null)`. Or at least marked with `@Nullable`.

src/main/java/com/palantir/gradle/gitversion/CommonGitOperations.java
```diff
@@ -0,0 +1,173 @@
...
+                return 0;
+            }
+
+            Matcher match = GIT_DESCRIBE_PATTERN.matcher(description);
+            Preconditions.checkState(match.matches(), "Cannot get commit distance for description: '%s'", description);
+            return Integer.parseInt(match.group(2));
+        }
+
+        private boolean isPlainTag() {
+            return !Pattern.matches(".*g.?[0-9a-fA-F]{3,}", description);
```
CRogers: This will compile the pattern every time right? We should probably have a static final `Pattern` in case it is called multiple times.

readme.md
```diff
@@ -100,27 +100,160 @@ Tasks
...
+```
+
+This will flow into publishing and a bunch of other builds. Sometimes you might not want to commit, other times you might need to pick a particular version. For whatever reason, you can do so via specifying `GIT_VERSION` as an environment variable when invoking Gradle:
+
+```
+> ./gradlew printVersion 
+1.0.0
+
+> GRADLE_VERSION=999 ./gradlew printVersion
+999/
```
CRogers: Trailing `/`

readme.md
```diff
@@ -100,27 +100,160 @@ Tasks
...
+        
+        // use this somehow
+    }
+} 
+```
+
+All calls are coalesced and run once for a given git root directory i.e. `<path>/.git`. So you can create as many instances and invoke as many times as required without taking a performance penalty.
+
+### Handling errors yourself
+This `Provider` will throw on resolution if the underlying `git` call fails. If you want finer control of the result, e.g. you want to handle a particular error, you can use `GitInvoker#invokeWithResult`:
```
CRogers: I would avoid `resolution` - too overloaded a term. I think "throw on read" is a better phrasing, more inline with the Gradle docs too.

readme.md
```diff
@@ -100,27 +100,160 @@ Tasks
...
+        CommonGitOperations commonGitOperations = getCommonGitOperationsFactory().create(parameters -> {
+           parameters.getPrefix().set("my-prefix@"); 
+        });
+        Provider<String> branch = commonGitOperations.branchName();
+        // use this somehow
+    }
+} 
+```
+
+### Making your own domain specific operations type
```
CRogers, line 206: I like

src/main/java/com/palantir/gradle/gitversion/GitExecOutput.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import org.immutables.value.Value;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Value.Immutable
+public abstract class GitExecOutput {
+
+    private static final Logger log = LoggerFactory.getLogger(GitExecOutput.class);
```
CRogers: `Logging.getLogger` - surprised the errorprone didn't complain...

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
```diff
@@ -16,118 +16,49 @@
...
+        return commonGitOperations.abbreviatedGitHash().get();
     }
 
     @Override
     public String getGitHashFull() throws IOException {
         try {
-            return gitHashFull.get();
+            return commonGitOperations.fullGitHash().get();
         } catch (RuntimeException e) {
             log.error("VersionDetailsImpl::getGitHashFull failed", e);
```
CRogers, line 63: I know this is from before - but lolwhut

CRogers: Just seems so strange to just log the error rather than keep the exception

CRogers:
LGTM apart from those comments

src/main/java/com/palantir/gradle/gitversion/VersionDetailsImpl.java
felixdesouza: lol yeah, when I tell you that `VersionDetailsImpl` is just meh as an api

src/main/java/com/palantir/gradle/gitversion/GitExecOutput.java
felixdesouza: I think maybe it's the `LoggerFactory` which is necessarily an slf4j type? but idk, I don't even check for that anymore because I assume errorprone will shout at me

CRogers (approved):

